PR #54, opened 2025-08-16
Twenty Icons favicon provider

ThatOneCalculator:
Adds [twenty icons](https://github.com/twentyhq/favicon) as a favicon provider, as a more privacy-friendly alternative to Google.


ThatOneCalculator:
@aurelien-brabant should be ready for review now :)

aurelleb:
LGTM, now I'm wondering whether we should set this as the default instead of google. I didn't know about twentyicons so I'm not sure how reliable it is? @ThatOneCalculator, what do you think?

ThatOneCalculator:
I think it'd be fine as the default, it's been live for 2 years w/o any problems. It is a relatively young project though, so I'll leave that final decision to you :)

aurelleb:
I think it's fine to set it as the default then. Thank you for your contribution, we really needed this!

Issue #67, opened 2025-08-17
Look for themes at XDG_DATA_DIRS locations


aurelleb:
Instead of only expecting themes to be put under `~/.config/vicinae/themes`, we should also look into system directories and, for the same id, prefer the local version.

Package maintainers will not have to symlink `/usr/share/vicinae/themes` anymore.

PR #71, opened 2025-08-17
chore: Added system directory for themes

cilginc:
Added system theme directory

### DIFF
--- a/vicinae/include/theme.hpp
+++ b/vicinae/include/theme.hpp
@@ -227,6 +227,7 @@ class ThemeService : public QObject {
   std::filesystem::path m_configDir;
   std::filesystem::path m_userThemeDir;
   std::filesystem::path m_dataThemeDir;
+  std::filesystem::path m_systemThemeDir;
 
   ThemeService(const ThemeService &rhs) = delete;
   ThemeService &operator=(const ThemeService &rhs) = delete;
--- a/vicinae/src/theme.cpp
+++ b/vicinae/src/theme.cpp
@@ -438,7 +438,11 @@ void ThemeService::upsertTheme(const ParsedThemeData &data) {
   *it = ThemeInfo::fromParsed(data);
 }
 
-void ThemeService::scanThemeDirectories() { scanThemeDirectory(m_userThemeDir); }
+void ThemeService::scanThemeDirectories() {
+  scanThemeDirectory(m_userThemeDir);
+  scanThemeDirectory(m_dataThemeDir);
+  scanThemeDirectory(m_systemThemeDir);
+}
 
 void ThemeService::scanThemeDirectory(const std::filesystem::path &path) {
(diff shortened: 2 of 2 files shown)


aurelleb:
I think we need to make this follow the `XDG_DATA_DIRS` paths, that would make more sense as the themes won't necessarily be installed in `/usr/share/vicinae` if the install prefix is different. See #67

cilginc:
> I think we need to make this follow the `XDG_DATA_DIRS` paths. That would make more sense, as the themes won't necessarily be installed in `/usr/share/vicinae` if the install prefix is different. See #67.

Firstly, I'm not that experienced in C++.  
Secondly, we also need to change `XDG_CONFIG_HOME` instead of using `.config` if you want to fully follow XDG standards.

Given the current state of the project, this PR is fine. We can migrate to using XDG environment variables later.

aurelleb:
> > I think we need to make this follow the `XDG_DATA_DIRS` paths. That would make more sense, as the themes won't necessarily be installed in `/usr/share/vicinae` if the install prefix is different. See #67.
> 
> Firstly, I'm not that experienced in C++. Secondly, we also need to change `XDG_CONFIG_HOME` instead of using `.config` if you want to fully follow XDG standards.
> 
> Given the current state of the project, this PR is fine. We can migrate to using XDG environment variables later.

Makes sense. We are not at risk of breaking anything here so let's merge this (let me test before)

cilginc:
> > > I think we need to make this follow the `XDG_DATA_DIRS` paths. That would make more sense, as the themes won't necessarily be installed in `/usr/share/vicinae` if the install prefix is different. See #67.
> > 
> > 
> > Firstly, I'm not that experienced in C++. Secondly, we also need to change `XDG_CONFIG_HOME` instead of using `.config` if you want to fully follow XDG standards.
> > Given the current state of the project, this PR is fine. We can migrate to using XDG environment variables later.
> 
> Makes sense. We are not at risk of breaking anything here so let's merge this (let me test before)

Of course do you want me to write some tests for PR's.

aurelleb:
> > > > I think we need to make this follow the `XDG_DATA_DIRS` paths. That would make more sense, as the themes won't necessarily be installed in `/usr/share/vicinae` if the install prefix is different. See #67.
> > > 
> > > 
> > > Firstly, I'm not that experienced in C++. Secondly, we also need to change `XDG_CONFIG_HOME` instead of using `.config` if you want to fully follow XDG standards.
> > > Given the current state of the project, this PR is fine. We can migrate to using XDG environment variables later.
> > 
> > 
> > Makes sense. We are not at risk of breaking anything here so let's merge this (let me test before)
> 
> Of course do you want me to write some tests for PR's.

no that's fine, I just like to test it locally to make sure it doesn't break anything.
but it's all good, merging. And thank you!

PR #80, opened 2025-08-18
feat: vim-like binds added

cilginc:
Edit: I changed my mind about adding toggle for vim binds.


aurelleb:
Can you add support for the grid-view or will that be a future PR?

cilginc:
> Can you add support for the grid-view, or will that be a future PR?

I didn’t add this as an extension, just as key bindings.  
A proper Vim extension would require a lot more functionality than I can handle right now.

aurelleb:
No I mean, there is a list view for strictly vertical lists, but there is also a grid view that is used to render grids (such as the one in the emoji command). It would make sense to have these "vim arrow key" bindings in both of them: https://github.com/vicinaehq/vicinae/blob/4ed9ced6e609ba0c3f53056cbf3b52a958ead749/vicinae/src/ui/views/grid-view.cpp#L9

Otherwise users might start wondering why the behaviour is inconsistent between the two

cilginc:
> No I mean, there is a list view for strictly vertical lists, but there is also a grid view that is used to render grids (such as the one in the emoji command). It would make sense to have these "vim arrow key" bindings in both of them:
> 
> https://github.com/vicinaehq/vicinae/blob/4ed9ced6e609ba0c3f53056cbf3b52a958ead749/vicinae/src/ui/views/grid-view.cpp#L9
> 
> Otherwise users might start wondering why the behaviour is inconsistent between the two

Oh i didn't see that i will update the code for gridviews too.

aurelleb:
Maybe you also want to add the same for the action panel: https://github.com/vicinaehq/vicinae/blob/4ed9ced6e609ba0c3f53056cbf3b52a958ead749/vicinae/src/action-panel/action-panel.hpp#L140

(I'm picky I know)

aurelleb:
Later on we can factorize this so that we have a centralized way of handling directional input but that will do for now.

cilginc:
I didn't tested it but it should work.

aurelleb:
thank you I will test it shortly

cilginc:
> thank you I will test it shortly

I tested it myself, and the clipboard history and custom plugins don’t work (I only tested the GIF search).  

I’ve identified the parts that don’t work and will push a fix soon.

aurelleb:
> > thank you I will test it shortly
> 
> I tested it myself, and the clipboard history and custom plugins don’t work (I only tested the GIF search).
> 
> I’ve identified the parts that don’t work and will push a fix soon.

Yeah I forgot to mention, the extensions use separate view types so you also have to change it there.

cilginc:
Todo fix these lines:

- [ ] https://github.com/vicinaehq/vicinae/blob/4ed9ced6e609ba0c3f53056cbf3b52a958ead749/vicinae/include/clipboard-history-view.hpp#L615C1-L627C6
- [ ] https://github.com/vicinaehq/vicinae/blob/4ed9ced6e609ba0c3f53056cbf3b52a958ead749/vicinae/src/ui/omni-list/omni-list.cpp#L736C1-L760C4
- [ ] https://github.com/vicinaehq/vicinae/blob/4ed9ced6e609ba0c3f53056cbf3b52a958ead749/vicinae/src/ui/action-pannel/action-pannel-list-view.cpp#L6C1-L19C4
- [ ] https://github.com/vicinaehq/vicinae/blob/4ed9ced6e609ba0c3f53056cbf3b52a958ead749/vicinae/src/extension/extension-grid-component.cpp#L11C1-L27C2
- [ ] https://github.com/vicinaehq/vicinae/blob/4ed9ced6e609ba0c3f53056cbf3b52a958ead749/vicinae/include/extension/extension-list-component.hpp#L176C1-L182C6

This files should be the files i need.

aurelleb:
You probably don't need to change `omni-list.cpp`, those handlers are not used anymore.

cilginc:
Everything works now.

Issue #81, opened 2025-08-19
feat: automatically launch/kill daemon


ArjixWasTaken:
## About

Currently, the expected way of using vicinae is to first run in the background `vicinae server`.
Otherwise, one could copy the systemd `.service` file to their user services and enable/start it using `systemctl --user enable --now vicinae`.

That is not ideal for various reasons:

1. The server running in the service may be of an older version, requiring user intervention in order to restart and use the newer bin.
2. Having a systemd service, implies there should also be support for other init systems.
3. It takes multiple steps to link/enable/start the systemd service, and there will be a plethora of issues opened related to that.
4. It's just not a smooth experience.

## Suggestion

Whenever vicinae is ran, it already checks for the presence of a daemon, you can make it automatically start the daemon when not present.

## Additional

You can also add a way for the daemon to identify its version number, so the client can kill it and restart it if there is a version mismatch.
(will definitely happen after a package update)

ArjixWasTaken:
If a user wants to run vicinae ahead of time (e.g. for cache to be built or whatever), they can utilize a one-off systemd unit, that only gets the daemon running and does not manage it.

ArjixWasTaken:
And this should not cause an issue if someone wants to see server logs, since running `vicinae server` kills the existing server either way, and they can observe the logs in the newly launched server.

ArjixWasTaken:
I can work on the PR if you want me to.

cilginc:
> I can work on the PR if you want me to.

I already maked a PR just wait for a moment.

ArjixWasTaken:
> > I can work on the PR if you want me to.
> 
> I already maked a PR just wait for a moment.

This?

<img width="1892" height="331" alt="Image" src="https://github.com/user-attachments/assets/3ee04566-a80f-4066-ba86-9d0e831667b7" />

cilginc:
> > > I can work on the PR if you want me to.
> > 
> > 
> > I already maked a PR just wait for a moment.
> 
> This?
[...]

PR #83, opened 2025-08-19
feat: added autostart server when running vicinae

cilginc:
See: https://github.com/vicinaehq/vicinae/issues/81
See: https://github.com/vicinaehq/vicinae/pull/82

I prefer this https://github.com/vicinaehq/vicinae/pull/82 one

### DIFF
--- a/vicinae/src/main.cpp
+++ b/vicinae/src/main.cpp
@@ -234,15 +234,23 @@ int main(int argc, char **argv) {
 
   DaemonIpcClient daemonClient;
 
-  if (!daemonClient.connect()) {
-    qInfo() << "Vicinae server not running. Starting it now...";
-    if (!QProcess::startDetached(qapp.applicationFilePath(), {"server"})) {
-      qCritical() << "Failed to start vicinae server.";
-      return 1;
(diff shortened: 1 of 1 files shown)


ArjixWasTaken:
Would having an env var that enables/disables this behavior be good?

PS: It being opt-in means it won't be the default behavior, and it also means you cannot reject it for going against any unix/linux philosophy that may cross your mind.

cilginc:
> Would having an env var that enables/disables this behaviour be good?

This PR probably wouldn't merge. I make this PR for you if you wanted to fork the project or compile with this PR.
You can add env vars for enabling this behavior of course.

ArjixWasTaken:
> This PR probably wouldn't merge.

And I am asking if making this opt-in would change that outcome.

cilginc:
> > This PR probably wouldn't merge.
> 
> And I am asking if making this opt-in would change that outcome.

Yes, this could be opt in. Which envvar should i choose for enabling this.

cilginc:
If you use `VICINAE_AUTO_START` env var it wil autostart the daemon now. Compile yourself and test it.

PR #90, opened 2025-08-20
fix systemd service

ArjixWasTaken:
I think you accidentally copy pasted an incorrect WantedBy to the service file.

user services cannot be wanted by multi-user, only system services can
as a result, systemd complains with a warning that it fails to find the unit multi-user

for now I am temporarily [patching this in my package,](https://git.arjix.dev/aur/vicinae-git/src/branch/master/PKGBUILD#L56) but I hope this change can be upstreamed

### DIFF
--- a/extra/vicinae.service
+++ b/extra/vicinae.service
@@ -12,4 +12,4 @@ Restart=always
 RestartSec=5
 
 [Install]
-WantedBy=multi-user.target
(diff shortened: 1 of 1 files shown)


aurelleb:
Yes, the very incompetent gpt-5 did that (good thing I don't use them to do actual coding!)

PR #102, opened 2025-08-21
fix(clipboard): Choose preferred mimeType for copy selection

RolfKoenders:
**UPDATE**

With less code I was trying to just choose the preferred mime type to show the copied text as discussed in the reactions on this merge request, which seems like the better way to do. 

I now have the following working.

image

<hr />

<details>
<summary>Original issue</summary>

When you now try to copy text from a firefox/zen browser window it shows up as text and not as a bunch of null characters. _See screenshot_

image

The selection text in the screenshot is what is being shown in the clipboard history. The output of `wl-paste` is the following 

```bash
text/html
text/_moz_htmlcontext
text/_moz_htmlinfo
text/plain;charset=utf-8
UTF8_STRING
COMPOUND_TEXT
TEXT
text/plain
STRING
text/plain;charset=utf-8
text/plain
text/x-moz-url-priv
```

I changed the priority of formats since it struggled with saving multiple versions of the same text. Its my first C++ contribution so please let me know or change if this is not needed and can be done differently. 
</details>

### DIFF
--- a/vicinae/src/services/clipboard/clipboard-service.cpp
+++ b/vicinae/src/services/clipboard/clipboard-service.cpp
@@ -8,6 +8,7 @@
 #include <qt6keychain/keychain.h>
 #include <QtConcurrent/QtConcurrent>
 #include <QFutureWatcher>
+#include <QTextDocument>
 #include "clipboard-server-factory.hpp"
 #include "crypto.hpp"
 #include "services/app-service/app-service.hpp"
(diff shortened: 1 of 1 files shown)


aurelleb:
It would help if you could explain with more precision what the issue is. Provide screenshots or a small video of you copying the text and having the unexpected behavior in the clipboard history. I'm not 100% sure I understand what the exact problem is (I can't repro on Firefox, didn't try Zen yet)

cilginc:
This also happens to me when using the Zen browser.  
If you select text in the browser and copy it, the clipboard includes the full HTML.  

- Pasting into a web input field is fine.  
- But if you paste into something like Vim, you’ll get output like this:
resim

aurelleb:
@cilginc is that after a direct paste or after your copy it from vicinae only?

cilginc:
> @cilginc is that after a direct paste or after your copy it from vicinae only?

After copy. 
Before copy it looks like this:
resim

aurelleb:
So the problem is that we are giving priority to the wrong mime type. We want the plain/text offer to be the one presented as the default for the selection (== the one shown in the history, and the one we generate the text preview from).
That said the fact it renders invalid glyphs is interesting, there might be another issue somewhere.

RolfKoenders:
I updated the description to reflect the current code change. Its a cleaner solution.

aurelleb:
> I updated the description to reflect the current code change. Its a cleaner solution.

Thank you this looks much better, I added a few comments, but it looks good. I will test it on my end today.

RolfKoenders:
Thanks for the comments! The code now indexes and stores all offers, including empty ones. The primary offer selection skips empty offers, so the history and preview always show meaningful content. 
Offer data is no longer mutated, only the preview extracts text from HTML for display. 
And I used your c++ tips!

aurelleb:
thank you! I will test this and merge it if I can't see anything wrong it

PR #103, opened 2025-08-21
feat: using ccache for build cache

cilginc:
I’ve noticed that the current workflow caching is a bit unstable,  
so I thought implementing **ccache** would be a great improvement.


cilginc:
Really??
resim

aurelleb:
> Really?? resim

That's it, Microsoft is ruining GitHub for good :rofl:

aurelleb:
> I’ve noticed that the current workflow caching is a bit unstable, so I thought implementing **ccache** would be a great improvement.

also note that  I added support for ccache in cmake very recently, I don't think it's as good as what you can do in the CI but it uses the system ccache install if available.

cilginc:
Services still not available which is crazy.

cilginc:
Yes, it was the problematic version of the GitHub Action, which is why it didn’t work.

Here's the new version. It is much more stable than the previous one. The old version took anywhere from 30 seconds to 2 minutes if you were lucky, and up to 4 minutes if you weren’t.

resim

The new version takes around 1 minute minimum. It's not faster than before but it is more reliable.

cilginc:
@aurelien-brabant Could you also update the Dockerfile to include `ccache` in the installed packages list and update the Action runner?  

The `ccache` Action already installs `ccache`, but we don’t want to reinstall it in every workflow.  
I can make a PR for this as well if you’d like.

aurelleb:
> @aurelien-brabant Could you also update the Dockerfile to include `ccache` in the installed packages list and update the Action runner?
> 
> The `ccache` Action already installs `ccache`, but we don’t want to reinstall it in every workflow. I can make a PR for this as well if you’d like.

I will have a look at this PR soon thank you.
The action runner has been updated with `ccache` installed for a few days already, but it's not synced with the repo so I forgot to update the Dockerfile here

PR #119, opened 2025-08-23
Clipboard support for Gnome, resolves #59

### LINKED ISSUE #59: Any solution to make clipboard work on gnome
Is it possible to make it work on Gnome? Because currently it shoes the pause and resume button still doesn't persist any history.

dagimg-dot:
https://github.com/user-attachments/assets/6259fb72-1e8d-436c-931a-c9f6d532f497

This PR adds clipboard support for Gnome through a safe fallback clipboard server implementation that doesn't interfere with wlr since wlr doesn't work on Gnome.


aurelleb:
oh but wait, so the QClipboard signal fires on gnome even for apps that don't currently have focus? On wlr environments it only reacts to clipboard events from inside the QT app itself.

dagimg-dot:
> oh but wait, so the QClipboard signal fires on gnome even for apps that don't currently have focus? On wlr environments it only reacts to clipboard events from inside the QT app itself.

I went the long path and tried to implement the AbstractWindowManager as GnomeWindowManager but after long trial and error using extensions to expose the cliboard events through d-bus it didn't work so I researched and finally AI suggested this and it worked. 

but what do you mean by apps that don't currently have focus? don't I need focus to select something on an app?

aurelleb:
The `dataChanged` signal only fires when you focus the Vicinae window back after you selected something elsewhere. If you copy something but the Vicinae window is closed, and then you copy something else the previous selection was not saved. I don't know if it still triggers on gnome, but at least on wlr envs it doesn't.

Essentially the QT app (here, Vicinae) is not able to be notified that the clipboard data changed unless it gets focus.

I don't have a gnome env to test this is rn but I might spin one up myself.

dagimg-dot:
> The `dataChanged` signal only fires when you focus the Vicinae window back after you selected something elsewhere. If you copy something but the Vicinae window is closed, and then you copy something else the previous selection was not saved. I don't know if it still triggers on gnome, but at least on wlr envs it doesn't.

owww....got you got you..so if vicinae is closed then i copied 2 things on other apps then open vicinae the latest one only appears in the clipboard. You are right. This means I need to work on my implementation further. Thanks 👍🏾

dagimg-dot:
> The `dataChanged` signal only fires when you focus the Vicinae window back after you selected something elsewhere. If you copy something but the Vicinae window is closed, and then you copy something else the previous selection was not saved. I don't know if it still triggers on gnome, but at least on wlr envs it doesn't.
> 
> Essentially the QT app (here, Vicinae) is not able to be notified that the clipboard data changed unless it gets focus.
> 
> I don't have a gnome env to test this is rn but I might spin one up myself.

I am going the gnome extension path because that seems the only viable solution. what do you think? **vicinae-gnome-extension** that exposes clipboard events and also window management.

aurelleb:
> I am going the gnome extension path because that seems the only viable solution. what do you think? **vicinae-gnome-extension** that exposes clipboard events and also window management.

Not a gnome expert at all but from what I've heard extensions are one of the only ways to do this.

dagimg-dot:
> > I am going the gnome extension path because that seems the only viable solution. what do you think? **vicinae-gnome-extension** that exposes clipboard events and also window management.
> 
> Not a gnome expert at all but from what I've heard extensions are one of the only ways to do this.

## PR Summary: GNOME Clipboard Integration

### Overview
This PR replaces the Qt-based clipboard server with a GNOME-specific clipboard server, enabling seamless clipboard integration between the main vicinae application and the GNOME desktop environment through the vicinae-gnome-extension.

### Key Changes

#### 1. **Replaced Qt Clipboard Server with GNOME Clipboard Server**
- **Removed**: `qt-clipboard-server.hpp/cpp` - Generic Qt clipboard fallback
- **Added**: `gnome-clipboard-server.hpp/cpp` - GNOME-specific clipboard implementation
- **Updated**: `clipboard-service.cpp` to use the new GNOME server as primary (priority 20) for gnome environments only

#### 2. **GNOME Environment Detection & Priority System**
- **Priority 20**: `GnomeClipboardServer` - GNOME + extension available
- **Priority 15**: `WlrClipboardServer` - Wayland non-GNOME environments
- **Automatic fallback**: System gracefully degrades if GNOME extension isn't available

#### 3. **D-Bus Integration Architecture**
The GNOME clipboard server communicates with the vicinae-gnome-extension through D-Bus:

```cpp
// D-Bus interface configuration
static constexpr const char *DBUS_SERVICE = "org.gnome.Shell";
static constexpr const char *DBUS_PATH = "/org/gnome/Shell/Extensions/Clipboard";
static constexpr const char *DBUS_INTERFACE = "org.gnome.Shell.Extensions.Clipboard";
```

### How It Works

#### **1. Extension Detection**
- Checks for GNOME environment (`XDG_CURRENT_DESKTOP`, `GDMSESSION`)
- Verifies D-Bus session bus availability
- Tests extension availability by calling `GetClipboardMimeTypes()` method
- Provides helpful error messages if extension is missing

#### **2. Clipboard Monitoring**
- Connects to `ClipboardChanged` D-Bus signal from the extension
- Calls `ListenToClipboardChanges()` to start monitoring
- Receives clipboard data in real-time with metadata:
  - Content (base64 encoded for binary data)
  - MIME type
  - Source application
  - Timestamp
  - Content hash
  - File size

#### **3. Data Processing**
- **Text data**: Direct UTF-8 conversion
- **Binary data**: Base64 decoding with fallback handling
- **Images**: Supports both data URLs and plain base64
- **Multiple formats**: Automatically generates additional MIME type offers (e.g., HTML for rich text)

#### **4. Connection Management**
- Automatic reconnection attempts every 5 seconds on connection loss
- Graceful cleanup of D-Bus connections
- Error handling with detailed logging

### Technical Implementation Details

#### **Signal Handling**
```cpp
// D-Bus signal signature: (stssssts)
void handleClipboardChanged(
    const QString &content,           // Base64 encoded content
    qulonglong timestamp,            // Unix timestamp
    const QString &source,           // Source identifier
    const QString &mimeType,         // MIME type
    const QString &contentType,      // Content category
    const QString &contentHash,      // Content hash
    qulonglong size,                 // Content size in bytes
    const QString &sourceApp         // Source application name
);
```

### Installation Requirements

Users must install the vicinae-gnome-extension for full functionality:
```bash
# Extension installation instructions provided in error messages
# Available at: https://github.com/dagimg-dot/vicinae-gnome-extension
```

#### **4. Additional features through the gnome extension**

1. Block clipboard on specified applications (password managers or any other)
2. Close vicinae-launcher on focus-loss (it was very annoying to use without this)

aurelleb:
thank you very much I will review it shortly

aurelleb:
LGTM! That's big, thanks a lot for this. I haven't been able to test it myself (only checked that it doesn't create an issue with my existing environment) but I'm sure you did and other gnome users will be very glad to!

Issue #60, opened 2025-08-17
Allow setting custom indexing locations for file search


Vortriz:
If I understand correctly, the file search currently only searches the home directory. It would nice to be able to specify custom locations (eg - mounted partitions etc) for file search.

aurelleb:
Planned but not configurable yet indeed

aurelleb:
Implemented by #126

PR #126, opened 2025-08-24
feat: Custom search paths

quadratech188:
Addresses #60.

Currently we can't use a custom `settingsDetail()` in `FileExtension` because `ExtensionRootProvider` ignores it and uses `preferences()` instead, so I had to use `Preference::makeText` as a semicolon-separated list.

Screenshot:


aurelleb:
I'm positively surprised you were able to go through that messy part of the code ahah, good job. That's an okay way to handle it for now, later we will probably make it a file picker preference when it gets implemented.

PR #139, opened 2025-08-27
added pywal converter python script

UnknownCalculatedVariable:
users can simply use this script which will make a file in XDG_CONFIG/vicinae/themes/pywal-theme.json and update with pywal colors.

it is just a simple conversion script


aurelleb:
thanks!

PR #140, opened 2025-08-27
fix: Remove duplicates via ID, not by filename

quadratech188:
This required actually generating the IDs properly according to [the XDG Desktop Entry spec](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html).

### DIFF
--- a/vicinae/src/services/app-service/xdg/xdg-app-database.cpp
+++ b/vicinae/src/services/app-service/xdg/xdg-app-database.cpp
@@ -1,5 +1,6 @@
 #include "xdg-app-database.hpp"
 #include "vicinae.hpp"
+#include <exception>
 #include <filesystem>
 #include <qlogging.h>
 #include <qsettings.h>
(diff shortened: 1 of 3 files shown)


aurelleb:
the id change breaks shortcuts because the saved id is no longer valid.
But since this is what the spec expects it makes more sense to have it be like that.
I think it's okay for us to break shortcuts in the state things currently are.
Users simply have to edit a non working shortcut and reselect the app to fix it.

vicinae/src/services/app-service/xdg/xdg-app-database.cpp
```diff
@@ -345,35 +349,28 @@ AppPtr XdgAppDatabase::findByClass(const QString &name) const {
...
-    apps.push_back(entry);
-    appMap.insert({entry->id(), entry});
-
-    for (const auto &action : entry->actions()) {
-      appMap.insert({action->id(), action});
-    }
-  } catch (const std::exception &except) {
-    qWarning() << "Failed to parse app at" << path << except.what();
-    return false;
+void XdgAppDatabase::addDesktopFile(fs::path parentPath, fs::path childPath) {
```
aurelien-brabant, line 352: any reason to pass fs::path as a value and not as a const ref?

aurelien-brabant (approved):
I think at some point we should make this a separate sub library under the main vicinae monorepo and add some tests (not our top priority)

PR #153, opened 2025-08-28
fix: Include .desktop extension in desktop entry ID

quadratech188:
#140 would generate an ID of `foo-bar` for foo/bar.desktop, when it should be `foo-bar.desktop`.


quadratech188:
This also resolves shortcuts breaking in #140.
Hopefully I don't make this kind of mistake again...

PR #176, opened 2025-08-30
feat: add Nix flake

TomRomeo:
This pull request adds Nix flake configuration files for Vicinae thanks to contributions of multiple people (see #6). With these config files, Vicinae can be built with Nix using the following command:
```bash
nix build github:vicinaehq/vicinae
```

Alternatively, Vicinae can also be added as a home-manager module. This will create a systemd service for the Vicinae server if enabled.

A Github workflow was also added to automatically build Vicinae on every commit to:
1. serve as a sanity check, showing if something broke the nix build process
2. push the created binaries to a binary cache (Cachix)

> [!WARNING]
> The Cachix action was configured for my fork. For the official Vicinae upstream, an official Cachix instance [should be created](https://docs.cachix.org/getting-started). If the url and name for the cache is provided, I can adapt the current workflow to reference the official Cachix instance.

### DIFF
--- a/.github/workflows/cachix.yaml
+++ b/.github/workflows/cachix.yaml
@@ -1,0 +1,22 @@
+name: "Push to Cachix"
+on:
+  pull_request:
+  push:
+
+jobs:
+  push_to_cachix:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


aurelleb:
Thank you, the long awaited flake is finally here :)

PR #189, opened 2025-08-31
refactor: Clean up file scanners

quadratech188:
Create the `ScanDispatcher` class to call the appropriate scanner - currently only `IncrementalScanner` and `IndexerScanner`, but easily extensible. It also manages the threads for these scanners.
This required creating `AbstractScanner` that all scanners derive from.
(I'm not completely satisfied about how both waiting for scans and scanning is in the same class, though)

Also moved `WriterWorker`, `Scan`, and `ScanType` to separate headers.


aurelleb:
A much needed refactor thanks a lot. I tested it and it looks like it's still working just fine.

PR #191, opened 2025-09-01
Fix Flatpak application icons not displaying in switch windows view on Gnome

dagimg-dot:
## Problem

Flatpak applications were showing generic document icons instead of their proper application icons in the switch windows view, while working correctly on the main page.

## Cause

A breaking change introduced in commit 4f2c472a that modified how application IDs are generated - switching from simple filenames to complex path-based IDs. The GNOME window manager, written before this change, still provides simple IDs, creating a mismatch that prevented Flatpak applications from being found in the app database.

## Solution

Enhanced the GNOME window manager to map simple Flatpak IDs to their corresponding complex path-based IDs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,13 +27,13 @@ Vicinae is designed for developers and power users who want fast, keyboard-first
 
 ## ✏️ Features
 
-> ⚠️ **Note:** Some features may vary depending on your desktop environment.  
-> If something isn’t supported yet, contributions are fully welcome — many integrations are beginner-friendly, and helpful guides are available in the [docs](https://docs.vicinae.com).
+> [!NOTE]
+> Some features may not be fully supported yet on some environments. 
 
 - Start and retrieve information about installed applications
 - File indexing with full-text search across millions of files — available via the file search module or directly from root search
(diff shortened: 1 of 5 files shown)


aurelleb:
@quadratech188 does this look okay to you?

quadratech188:
I think this should work fine, in that it retains original behavior.

But there's something I don't understand about the existing code: I'm familiar with neither flatpak nor GNOME, but I think it's fair to assume that `wm_class` refers to `StartupWMClass` in `.desktop` files - shouldn't we be matching for that, not the app id?

aurelleb:
Waiting for @dagimg-dot input on this as he is the gnome guy, but what I'm doing in some places in the app is that I try to match on the wm_class first and then if there is no match I try the id.

dagimg-dot:
> Waiting for @dagimg-dot input on this as he is the gnome guy, but what I'm doing in some places in the app is that I try to match on the wm_class first and then if there is no match I try the id.

I only changed the original implementation of gnome-window-manager to reflect @quadratech188 changes on the use of path-based ids.

aurelleb:
ok we will take this in as it is for now

PR #194, opened 2025-09-01
feat: Adding A Lot of Themes

cilginc:


aurelleb:
thanks! (that's indeed a lot of themes)

PR #197, opened 2025-09-02
Fix Extension Development Auto-Refresh UI

dagimg-dot:
### Problem
Extension development auto-refresh was working (rebuilding extensions), but the Vicinae launcher UI never showed the new change but after the first fix, this got solved. And after that it started showing visual corruption with overlapping text and distorted command lists when extensions were refreshed during development.

### Root Cause
Two issues:

1.`registerRepository()` didn't emit `registryAdded` signal when replacing existing repositories, so the UI never knew to refresh
2.  `RootExtensionManager` added new providers without removing old ones, causing duplicate entries and UI corruption

### Solution
**Two minimal fixes:**

1. emit `registryAdded(repository)` when replacing existing repositories

2. remove existing provider before adding new one:

Extension development workflow now works without restarting the vicinae server whenever the extension code changes or the manifest file is updated.

### DIFF
--- a/vicinae/include/omni-command-db.hpp
+++ b/vicinae/include/omni-command-db.hpp
@@ -102,6 +102,7 @@ public:
       if (repositories[i]->id() == repository->id()) {
         repositories[i] = repository;
         index = i;
+        emit registryAdded(repository);
         break;
       }
     }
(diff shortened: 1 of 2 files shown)


aurelleb:
Ok, I definitely need to come up with a better solution for this but that fixes it.

dagimg-dot:
> Ok, I definitely need to come up with a better solution for this but that fixes it.

Okay, but its such a relief to develop extensions with hot reload 😁

PR #198, opened 2025-09-02
fix: Nord theme icon path

cilginc:
This commit corrects the file path that was mistakenly set in the previous commit.


cilginc:
**Note:** There seems to be an issue with the Nix workflows. They shouldn’t be running for **4 hours**.

PR #201, opened 2025-09-02
Fix building on nix for good

bendi-github:
When I wanted to build vicinae from the provided nix flake earlier today, the build failed because of a hash mismatch. This happened because something about the npm deps got changed without modifying the hard-coded hashes in the nix build file (this is expected nix behavior). In a nix-centric project this issue would not occur, because they build via nix and catch these issues when they happen. Since the nix build was donated to this repo and you (lead dev who made the breaking change) are not using it, it gets out of sync. 

To solve this issue, this PR adds an update script, which updates the hashes nix uses and can be run via the provided GH action, if you (understandably so) do not want to use nix. This is my first time messing around with GH actions, thus my GH action might not be optimal.

I would have liked to fetch the npm deps as a flake input, but that does not seem to be possible, hence the dubious workarounds employed by me.

### DIFF
--- a/.github/workflows/nix-support.yml
+++ b/.github/workflows/nix-support.yml
@@ -1,0 +1,20 @@
+name: Nix hash check
+
+on:
+  push:
+  workflow_dispatch:
+
+
+jobs:
+  update-tags:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+      - uses: cachix/install-nix-action@v31
+      - name: Check nix hashes
+        run: |
+          nix run .#nix-update-script --extra-experimental-features nix-command flakes --accept-flake-config
--- a/flake.nix
+++ b/flake.nix
@@ -22,9 +22,24 @@
       let
         pkgs = nixpkgs.legacyPackages.${system};
         vicinaePkg = pkgs.callPackage ./vicinae.nix { };
+        nix-update-script = pkgs.writeShellScriptBin "nix-update-script" ''
+          OLD_API_DEPS_HASH=$(${pkgs.lib.getExe pkgs.nix} eval --raw .#packages.x86_64-linux.default.passthru.apiDeps.hash)
+          OLD_EXT_MAN_DEPS_HASH=$(${pkgs.lib.getExe pkgs.nix} eval --raw .#packages.x86_64-linux.default.passthru.extensionManagerDeps.hash)
+
+          cd api
+          NEW_API_DEPS_HASH=$(${pkgs.lib.getExe pkgs.prefetch-npm-deps} package-lock.json)
+          cd ../extension-manager
+          NEW_EXT_MAN_DEPS_HASH=$(${pkgs.lib.getExe pkgs.prefetch-npm-deps} package-lock.json)
+          cd ..
+
+          [[ "$OLD_API_DEPS_HASH" == "$NEW_API_DEPS_HASH" ]] || { echo -e "\e[31mHash mismatch for API npm deps, please replace the value in vicinae.nix with '$NEW_API_DEPS_HASH'.\e[0m" >&2; exit 1;}
+
+          [[ "$OLD_EXT_MAN_DEPS_HASH" == "$NEW_EXT_MAN_DEPS_HASH" ]] || { echo -e "\e[31mHash mismatch for extension-manager npm deps, please replace the value in vicinae.nix with '$NEW_EXT_MAN_DEPS_HASH'.\e[0m" >&2; exit 1;}
+        '';
       in
       {
         packages.default = vicinaePkg;
+        packages.nix-update-script = nix-update-script;
         devShells.default = pkgs.mkShell {
           inputsFrom = [ vicinaePkg ]; # automatically pulls nativeBuildInputs + buildInputs
           buildInputs = [
--- a/vicinae.nix
+++ b/vicinae.nix
@@ -157,4 +157,7 @@ stdenv.mkDerivation rec {
     license = licenses.gpl3Plus;
     mainProgram = "vicinae";
   };
+  passthru = {
+    inherit apiDeps extensionManagerDeps;
+  };
 }


aurelleb:
I'm wondering whether using a github action for this is the best way, wouldn't a simple git commit hook do the trick?
Also the conflict is because we just merged a quick fix updating the hash, no big deal.

bendi-github:
The github action only exists, because I didn't want to require a local installation of nix. It is possible to run `nix run .#nix-update-script --extra-experimental-features nix-command flakes --accept-flake-config` from a local CLI to achieve the same result. That command could very well be a git hook, but you would need a local installation of nix. When using nix to build things, the nix cache quickly grows. Garbage collection does not happen automatically, thus you would need to familiarize yourself with nix to not clog up your dev machine. I assumed pushing this responsibility onto you would not be very nice.

bendi-github:
I think the merge conflict stems from me renaming `vicinae.nix` to `package.nix` (which is weird, because locally, git recognizes the rename). I did this, because nixpkgs (nix' package repository) does it like that, with the intention of eventually upstreaming it there. I don't know whether this is desired (@TomRomeo what do you think?), but if there is interest I could look into further cleanup of the package definition.

Vortriz:
@bendi-github Please add `meta` for the package, including `meta.mainProgram = "vicinae"`. I would have made a PR but it's too small of a change.

TomRomeo:
Hi, sorry for the late reply. I would love if we could add this to Nixpgs! Maybe we can create a separate issue / discussion for this effort.

Regarding the Github Actions: I can't explain why but I dislike an action that would write commits to the git tree. However, I think adding this as a pre-commit hook is also not great for most contributors due to missing Nix tools as was outlined already.

What do you guys think about an action that does only the hash check for both folders and simply fails if the hashes are out of date? The action could simply display the correct hash and a small explanation where the hash would have to be adapted. This way it acts like a validation check (and doesn't take multiple hours in case of wrong hashes like the build action does atm). I think this would make fixing the hashes especially in PRs more easy (No merging if commit pipelines fail) and also shows the error and fix quickly if the code is comitted directly to the main branch.

aurelleb:
Yeah I think using the action as an alert that something is off can be good enough.

bendi-github:
Okay, I have reworked the code: The GHA now runs on every push and checks whether the hashes match. If they don't match, the action fails with a helpful error message. I have kept the `nix-support.json` file, because I couldn't think of a way to parse `vicinae.nix` reliably. One issue with this approach is that the flake lockfile `flake.lock` is never updated. This isn't that big of a deal, because it doesn't break builds.

TomRomeo:
Maybe you can try to use nix eval in the github action to avoid the extra json file? I have sketched an example GHA with ChatGPT below but could not try it out yet.

```yaml

name: Check npmDepsHash correctness

on:
  push:
    branches:
      - '**'   # run on every branch push

jobs:
  check-npmdepshash:
    runs-on: ubuntu-latest

    steps:
      - name: Checkout repo
        uses: actions/checkout@v4

      - name: Install Nix
        uses: cachix/install-nix-action@v27

      - name: Build prefetch-npm-deps
        run: nix build nixpkgs#prefetch-npm-deps

      - name: Run prefetch-npm-deps for ./api
        run: |
          ./result/bin/prefetch-npm-deps ./api/package-lock.json > api.hash
          echo "API_HASH=$(cat api.hash)" >> $GITHUB_ENV

      - name: Run prefetch-npm-deps for ./extension-manager
        run: |
          ./result/bin/prefetch-npm-deps ./extension-manager/package-lock.json > extension-manager.hash
          echo "EXT_HASH=$(cat extension-manager.hash)" >> $GITHUB_ENV

      - name: Extract npmDepsHash from vicinae.nix
        run: |
          API_NIX_HASH=$(nix eval --raw -f vicinae.nix apiDepsHash)
          EXT_NIX_HASH=$(nix eval --raw -f vicinae.nix extensionManagerDepsHash)
          echo "API_NIX_HASH=$API_NIX_HASH" >> $GITHUB_ENV
          echo "EXT_NIX_HASH=$EXT_NIX_HASH" >> $GITHUB_ENV

      - name: Compare hashes
        run: |
          FAILED=0

          echo "API hash (prefetch):        $API_HASH"
          echo "API hash (vicinae.nix):     $API_NIX_HASH"
          if [ "$API_HASH" != "$API_NIX_HASH" ]; then
            echo "❌ Error: apiDepsHash mismatch"
            FAILED=1
          else
            echo "✅ apiDepsHash matches"
          fi

          echo ""
          echo "Extension hash (prefetch):  $EXT_HASH"
          echo "Extension hash (vicinae.nix): $EXT_NIX_HASH"
          if [ "$EXT_HASH" != "$EXT_NIX_HASH" ]; then
            echo "❌ Error: extensionManagerDepsHash mismatch"
            FAILED=1
          else
            echo "✅ extensionManagerDepsHash matches"
          fi

          exit $FAILED
```

bendi-github:
LLMs and nix don't tend to mix well in my experience, which is also true in this case. `nix eval` just doesn't work how ChatGPT would like it to work. You can't just eval a package definition and expect it to work (the command fails with an error). One could try to call `callPackage` on the file contents and then try to figure out how to get the hash. I inspected the resulting derivation in a repl, but couldn't find the hashes, as I expected. What makes this especially difficult I think is the `let`-binding used for fetching the deps. This makes it nearly impossible to get at via the repl (and thus also via the cli). If the derivation contained a bog-standard `src` attribute for example, the override mechanisms would allow us to get/set that value. `let` bindings do not afford us such luxuries, but I should mention that I am far from an expert in the nix language. 

What could work is passing the hashes as parameters to `vicinae.nix` like
```nix
...,
apiDepsHash ? "<hash>",
extensionDepsHash ? "<hash>",
...
```
and then just adjusting the default parameter. That way would work I think, but at that point parsing could just be done via `awk`.

TomRomeo:
You are right, I should have tested the solution first before posting. (It just seemed like it _should_ work but it doesn't 🥲)
After fiddling around with this a little bit, it seems like we could simply inherit apiDeps and extensionManagerDeps in the derivation to access them from nix eval though or we pass them to the [passthru](https://nixos.org/manual/nixpkgs/stable/#chap-passthru) attribute.

### Example:
```nix
# vicinae.nix
{ ... }:
let
  src = ./.;

  # Prepare node_modules for api folder
  apiDeps = fetchNpmDeps { ... };

  # Prepare node_modules for extension-manager folder
  extensionManagerDeps = fetchNpmDeps { ... };

in
  stdenv.mkDerivation rec {
    name = "vicinae";

    inherit src;

    passthru = {
      inherit apiDeps extensionManagerDeps;
    };

    nativeBuildInputs = [
		"..."
    ];

    buildInputs = [
		"..."
    ];

}
```
```bash
nix eval --raw .#packages.x86_64-linux.default.passthru.apiDeps.hash
```

This works for me when testing it but maybe this solution is a bit too hacky? What do you think?

bendi-github:
I think what you proposed is a good solution. That way we don't have to rely on hacky file parsing. The only complaint one could have is that cursory knowledge of nix is required to change the hashes. I have adjusted my commit accordingly.

schromp:
Is it not a bit early for upstreaming to nix? We have a lot of version changes atm and i fear that the version in nixpkgs will be outdated fast.

As far as i know it is prefered to have projects with rapid development (especially at the start of the project) outside of nixpkgs and upstream later when the versions get more stable and you wont have to open a pr every 2 days to bump the version.



> Hi, sorry for the late reply. I would love if we could add this to Nixpgs! Maybe we can create a separate issue / discussion for this effort.

TomRomeo:
> Is it not a bit early for upstreaming to nix? We have a lot of version changes atm and i fear that the version in nixpkgs will be outdated fast.
> 
> As far as i know it is prefered to have projects with rapid development (especially at the start of the project) outside of nixpkgs and upstream later when the versions get more stable and you wont have to open a pr every 2 days to bump the version.
> 
> > Hi, sorry for the late reply. I would love if we could add this to Nixpgs! Maybe we can create a separate issue / discussion for this effort.

You are probably right. With the overlay that we have in the flake at the moment, we can wait until the development might have slowed down a bit :)

aurelleb:
I think what I just merged introduced a small conflict. Once that last one is resolved I think this is ready to merge?

PR #207, opened 2025-09-03
feat: Implement WatcherScanner

quadratech188:
Uses [e-dant/watcher](https://github.com/e-dant/watcher) to watch for changes.

Very half-baked, but I wanted to get it out first.

Stuff to work out before merge:
- How to include `watcher.hpp`: ([watcher](https://github.com/e-dant/watcher) is a header-only library)
As of now I just pasted the file, but I think there's a way to do it using some CMake magic.

- How to make this optional:
This is probably *very* buggy, so I need to make this optional.
I went to add a setting to `FileExtension`, but the code there uses `AbstractFileIndexer` - and putting a `useWatcher()` function in that abstract class just isn't pretty.

### DIFF
--- a/vicinae/CMakeLists.txt
+++ b/vicinae/CMakeLists.txt
@@ -436,6 +436,7 @@ set(SRCS
 	src/services/files-service/file-indexer/relevancy-scorer.cpp
 	src/services/files-service/file-indexer/incremental-scanner.cpp
 	src/services/files-service/file-indexer/indexer-scanner.cpp
+	src/services/files-service/file-indexer/watcher-scanner.cpp
 	src/services/files-service/file-indexer/home-directory-watcher.cpp
 	src/services/files-service/file-indexer/file-indexer-db.cpp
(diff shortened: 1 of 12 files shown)


aurelleb:
any reason not to use https://doc.qt.io/qt-6/qfilesystemwatcher.html ?

quadratech188:
It doesn't watch subdirectories.

aurelleb:
also unless it's well packaged pasting the file is fine, that's how it's done.

quadratech188:
I spent some time refactoring, I think this is ready enough now.

aurelleb:
the app image build fails on the std::ranges::to usage

aurelleb:
ideally I should find a way to compile a working version of modern gcc on the app image builder (which runs ubuntu-22), but last time I tried I failed

quadratech188:
Build succeeded.

aurelleb:
perfect then, thank you!

PR #220, opened 2025-09-04
feat: Add control abilities to AbstractScanner and ScanDispatcher

quadratech188:
Needed for further development on #207.


quadratech188:
Hang on, the scanners aren't running properly

PR #225, opened 2025-09-05
Windows Navigation: generic support for compositors that support the extension wlr-foreign-toplevel-management-unstable-v1

dawsers:
This works for most wlroots based compositors and probably Hyprland too, even though vicinae already has a special extension for it using IPC calls.

I took most of the code from `rofi` and adapted it. It is MIT licensed.

Main changes:

1. Changed a couple of small things in the build system to reuse the `wayland-scanner` function to generate protocol files.
2. Made `WindowManager::start() `non `const`, as it is a good place to initialize data, for example in this case. Most other managers are not using it, because they rely on re-doing the list of windows every time it is requested, but the implementation in this PR creates a listener and only updates the window list when needed.
3. Call `WindowManager::start()` which wasn't being done. This proves the function wasn't getting much love :-)
4. Implement `wayland.hpp` and `wayland.cpp` re-using a lot of code from `rofi`'s wayland windows mode.
5. Add the `xml` file for the protocol under `vicinae/protocols/wlr-foreign-toplevel-management-unstable-v1.xml`

### DIFF
--- a/vicinae/src/services/window-manager/wayland/wayland.cpp
+++ b/vicinae/src/services/window-manager/wayland/wayland.cpp
@@ -30,13 +30,21 @@ static void foreign_toplevel_handle_output_leave(void *data,
 
 static void foreign_toplevel_handle_state(void *data,
     struct zwlr_foreign_toplevel_handle_v1 *handle, struct wl_array *value) {
-  // ignore
+  WaylandWindow *self = static_cast<WaylandWindow *>(data);
+
+  self->m_active = false;
+  size_t n = value->size / sizeof(uint32_t);
+  for (size_t i = 0; i < n; ++i) {
+      uint32_t *elem = (uint32_t *)value->data + i;
+      if (*elem == ZWLR_FOREIGN_TOPLEVEL_HANDLE_V1_STATE_ACTIVATED) {
+          self->m_active = true;
+      }
+  }
 }
 
 static void foreign_toplevel_handle_done(void *data,
     struct zwlr_foreign_toplevel_handle_v1 *handle) {
   WaylandWindow *self = static_cast<WaylandWindow *>(data);
-
   emit self->m_manager->windowsChanged();
 }
 
@@ -51,7 +59,6 @@ static void foreign_toplevel_handle_closed(void *data,
   }
 
   // the handle is inert and will receive no further events
-  delete self;
   emit self->m_manager->windowsChanged();
 }
 
@@ -78,6 +85,7 @@ WaylandWindow::WaylandWindow(WaylandWindowManager *wm,
   m_manager = wm;
   m_handle = handle;
   m_pid = 0;
+  m_active = false;
   zwlr_foreign_toplevel_handle_v1_add_listener(handle,
     &foreign_toplevel_handle_listener, this);
 }
@@ -97,7 +105,13 @@ AbstractWindowManager::WindowList WaylandWindowManager::listWindowsSync() const 
 }
 
 AbstractWindowManager::WindowPtr WaylandWindowManager::getFocusedWindowSync() const {
-    return nullptr;
+  for (auto window : m_toplevels) {
+      WaylandWindow *ww = static_cast<WaylandWindow *>(window.get());
+      if (ww->m_active) {
+          return window;
+      }
+  }
+  return nullptr;
 }
 
 void WaylandWindowManager::focusWindowSync(const AbstractWindow &window) const {
@@ -137,7 +151,9 @@ static void foreign_toplevel_manager_toplevel(void *data,
 
 static void foreign_toplevel_manager_finished(void *data,
     struct zwlr_foreign_toplevel_manager_v1 *manager) {
+  WaylandWindowManager *wm = static_cast<WaylandWindowManager *>(data);
   zwlr_foreign_toplevel_manager_v1_destroy(manager);
+  wm->m_manager = nullptr;
 }
 
 static struct zwlr_foreign_toplevel_manager_v1_listener
--- a/vicinae/src/services/window-manager/wayland/wayland.hpp
+++ b/vicinae/src/services/window-manager/wayland/wayland.hpp
@@ -21,6 +21,7 @@ public:
   QString m_title;
   QString m_wmClass;
   int m_pid;
+  bool m_active;
   struct zwlr_foreign_toplevel_handle_v1 *m_handle;
   WaylandWindowManager *m_manager;
 };


aurelleb:
**Edit**: found what the issue is, you are deleting the window manually although tracking it with a shared pointer already (as part of the `m_toplevels` window list). So erasing the iterator triggers a first deletion and your delete is performing a double free. It's fixed by removing the manual deletion.

First of all, thank you for your work.

I don't know why I didn't think about this before, for some reason I made the assumption that doing native wayland stuff inside the QT app itself would cause issues with the main QT things (most notably the calls to `wayland_display_roundtrip`). But it looks like I was simply wrong here.

I also agree with the changes you made to the abstract window manager class.

Upon testing however, I quickly came across a few crashes. I'm attaching a gdb backtrace below.

It looks like it's the window deletion being at fault here.
A first issue is that you delete `self` to use it right after, but changing that didn't solve the crash. It looks like the crash has to do with the deletion itself. However you don't attempt to delete it elsewhere so it doesn't seem to be a double free issue.

My guess is that there is memory corruption happening somewhere else and causing this as a result (FYI, I'm on Hyprland)

I haven't had time to explore this further, but reproducing is very easy on my system:

- Open vicinae server
- Close any window
- Blows up

```
(gdb) bt 20
#0  0x0000555555cafb75 in std::__atomic_base<int>::fetch_sub (this=0x3e9a910249ae4c48, __i=1, __m=std::memory_order::acq_rel) at /usr/lib/gcc/x86_64-pc-linux-gnu/14/include/g++-v14/bits/atomic_base.h:641
#1  QAtomicOps<int>::deref<int> (_q_value=<error reading variable: Cannot access memory at address 0x3e9a910249ae4c48>) at /usr/include/qt6/QtCore/qatomic_cxx11.h:266
#2  0x0000555555c9a870 in QBasicAtomicInteger<int>::deref (this=0x3e9a910249ae4c48) at /usr/include/qt6/QtCore/qbasicatomic.h:48
#3  0x0000555555c751a6 in QArrayData::deref (this=0x3e9a910249ae4c48) at /usr/include/qt6/QtCore/qarraydata.h:66
#4  0x0000555555cb01df in QArrayDataPointer<char16_t>::deref (this=0x55555858f548) at /usr/include/qt6/QtCore/qarraydatapointer.h:452
#5  0x0000555555c9b42a in QArrayDataPointer<char16_t>::~QArrayDataPointer (this=0x55555858f548, __in_chrg=<optimized out>) at /usr/include/qt6/QtCore/qarraydatapointer.h:108
#6  0x0000555555c764f4 in QString::~QString (this=0x55555858f548, __in_chrg=<optimized out>) at /usr/include/qt6/QtCore/qstring.h:1411
#7  0x0000555556260111 in WaylandWindow::~WaylandWindow (this=0x55555858f540, __in_chrg=<optimized out>) at /home/aurelle/prog/perso/vicinaehq/vicinae/vicinae/src/services/window-manager/wayland/wayland.cpp:90
#8  0x000055555625ff61 in foreign_toplevel_handle_closed (data=0x55555858f540, handle=0x5555572e4440) at /home/aurelle/prog/perso/vicinaehq/vicinae/vicinae/src/services/window-manager/wayland/wayland.cpp:54
#9  0x00007ffff44a7056 in ??? () at /usr/lib64/libffi.so.8
#10 0x00007ffff44a5e7e in ??? () at /usr/lib64/libffi.so.8
#11 0x00007ffff44a66ab in ffi_call () at /usr/lib64/libffi.so.8
#12 0x00007ffff60e2a18 in wl_closure_invoke (closure=closure@entry=0x7fffdc0016d0, flags=flags@entry=1, target=<optimized out>, target@entry=0x5555572e4440, opcode=opcode@entry=6, data=<optimized out>) at ../wayland-1.24.0/src/connection.c:1241
#13 0x00007ffff60de514 in dispatch_event (display=display@entry=0x5555563bab40, queue=queue@entry=0x5555563bac38) at ../wayland-1.24.0/src/wayland-client.c:1707
#14 0x00007ffff60df6fb in dispatch_queue (display=0x5555563bab40, queue=0x5555563bac38) at ../wayland-1.24.0/src/wayland-client.c:1853
#15 wl_display_dispatch_queue_pending (display=0x5555563bab40, queue=0x5555563bac38) at ../wayland-1.24.0/src/wayland-client.c:2190
#16 0x00007ffff619ff32 in ??? () at /usr/lib64/libQt6WaylandClient.so.6
#17 0x00007ffff6180e41 in QtWaylandClient::QWaylandDisplay::flushRequests() () at /usr/lib64/libQt6WaylandClient.so.6
#18 0x00007ffff54b4910 in QObject::event (this=0x5555564e0730, e=0x7fffe4001d20) at /var/tmp/portage/dev-qt/qtbase-6.9.1-r3/work/qtbase-everywhere-src-6.9.1/src/corelib/kernel/qobject.cpp:1431
#19 0x00007ffff7d9d246 in QApplicationPrivate::notify_helper (this=<optimized out>, receiver=0x5555564e0730, e=0x7fffe4001d20) at /var/tmp/portage/dev-qt/qtbase-6.9.1-r3/work/qtbase-everywhere-src-6.9.1/src/widgets/kernel/qapplication.cpp:3303
```

dawsers:
It seems what I mentioned about atomic updates doesn't apply to closing windows, so just fixing what you found about double deleting the window (shared pointer + delete) should fix the issue. I also added `m_active` to know the focused window. It is not very useful at the moment, because the current state of the extension doesn't allow sending keyboard events to the window, but that may change through some other Wayland extension, so I will leave it there.

Thank you for the project and your help!

aurelleb:
Yeah and thank you for showing me that doing wayland stuff in QT directly is no issue at all, I will probably end up moving the clipboard stuff inside then.
And I'm glad you are liking the project!

Also, side note regarding focus: it works but not having it wm integrated makes it less useful because it's not going to handle automatic workspace switching (nothing you can do about that, just noting it)
I wish we could get the same protocol but applied to workspaces

dawsers:
> Also, side note regarding focus: it works but not having it wm integrated makes it less useful because it's not going to handle automatic workspace switching (nothing you can do about that, just noting it) I wish we could get the same protocol but applied to workspaces

I haven't been using Hyprland for a while, but if the compositor implements the extension correctly, it should change workspace too. I am using my own compositor,  [scroll](https://github.com/dawsers/scroll), and when I select a window and press `<cr>` it changes workspace if the window is in another workspace. scroll is a sway fork, so most wlroots compositors should behave the same way.

aurelleb:
Ah, I guess Hyprland is not implementing this properly then, it led me to think it wasn't working at first but it was only a workspace issue. Maybe it's worth reporting upstream? idk

dawsers:
I think Hyprland is closed to external issues. I used to contribute when I was a user and plugin developer, but then they closed the ability to file issues. I still have an open PR there, but I am no longer a user.

aurelleb:
Yeah Vaxry can't do it all after all. That won't matter to us anyways since the Hyprland implementation will be activated instead of the generic one.

PR #238, opened 2025-09-07
fix(nix): parse manifest.yaml for nix build

### LINKED ISSUE #212: VICINAE_GIT_TAG is wrong when building from Nix flake
Since it's built outside of the git tree, the only way to provide version and commit info is to inject it manually during cmake configuration.
Currently we are setting the hash as the version which is technically incorrect: https://github.com/vicinaehq/vicinae/blob/d092a6816d50d4b691d0c19815e9fceaa66f24aa/vicinae.nix#L53C1-L54C1

This can be problematic as  we use this in the extension boilerplate generation logic to set the version of the `@vicinae/api` package.

Image

TomRomeo:
This pull request adapts the Nix derivation to read version information from the manifest.yaml in the directory.
This resolves #212 .

SCR-20250907-qcfh

Turns out Nix does not have a native Yaml parsing function? So I implemented a whacky parser with `builtins.match` and a [regex](https://regex101.com/r/vgsgYY/1)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -15,7 +15,7 @@
     flake-utils.lib.eachDefaultSystem (system:
       let
         pkgs = nixpkgs.legacyPackages.${system};
-        vicinaePkg = pkgs.callPackage ./vicinae.nix { hash = self.rev ? self.dirtyRev; };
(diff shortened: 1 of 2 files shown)


aurelleb:
thanks!

PR #292, opened 2025-09-14
Run scans per-directory, not per-scanner

quadratech188:
An `AbstractScanner` now handles a single `Scan`, not an entire `ScanType`. `ScanDispatcher` manages the `AbstractScanner`s and colects them when the scans are finished.

Having multiple scans going simultaneously showed that `FileIndexerDatabase` (Or more accurately; SQLite) can't handle concurrent writes and often throws `database locked`.
So I implemented `DbWriter` that receives writes from all the scanners and queues them.

I noticed that `ntfs-3g`'s CPU usage doesn't die down for quite a while after the scans have finished: I attribute this to `fs::last_write_time` calls when calculating scores for each entry. So storing that info while scanning could be better for performance.


aurelleb:
Amazing work, thank you! This will probably fix many of our ongoing issues with the file indexing.

A recent PR I merged introduced some conflicts though, I had a look at it and it's nothing too big it seems. Feel free to discard what I did with the 'waiting for writing thread' thing.

Also I worked on making vicinae compile on slightly older systems, and in order to do so, had get rid of bleeding edge C++23 features so no `std::expected` anymore. 

I added `tl::expected` to the project instead, which is an implementation of expected that works the same so you can just use `tl::expected` instead.

quadratech188:
> This will probably fix many of our ongoing issues with the file indexing.

I bet it'll introduce far more...

The project compiling means we don't have any more C++23 bits, right?

aurelleb:
> > This will probably fix many of our ongoing issues with the file indexing.
> 
> I bet it'll introduce far more...
> 
> The project compiling means we don't have any more C++23 bits, right?

yes I got rid of some range stuff such as std::ranges::to

PR #294, opened 2025-09-14
Add `settings` option to home-manager module

kaan-w:
Allows users to declare `~/.config/vicinae/vicinae.json` using nix.

Example configuration:
```nix
services.vicinae = {
  enable = true;
  autoStart = true;

  # New
  settings = {
      faviconService = "twenty";
      font.size = 11;
      popToRootOnClose = false;
      rootSearch.searchFiles = false;
      theme.name = "vicinae-dark";
      window = {
        csd = true;
        opacity = 0.85;
        rounding = 10;
      };
   };
};
```

### DIFF
--- a/module.nix
+++ b/module.nix
@@ -69,12 +69,41 @@ in
           '';
       type = lib.types.attrsOf lib.types.attrs;
     };
+
+    settings = lib.mkOption {
+      type = lib.types.attrs;
+      default = { };
+      description = "Settings written as JSON to `~/.config/vicinae/vicinae.json.";
+      example = lib.literalExpression ''
+        {
+          faviconService = "twenty";
+          font = {
+            size = 10;
+          };
+          popToRootOnClose = false;
+          rootSearch = {
+            searchFiles = false;
+          };
+          theme = {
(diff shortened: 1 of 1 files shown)


aurelleb:
the `vicinae.json` configuration is not meant to be stable, eventually it will probably make it inside the database or we will change the configuration format. We can take this is in for now but eventually we will have to refactor this

aurelleb:
But I get the idea of being able to declaratively configure it. only if we do that we might as well make the switch to toml

wvy777:
I've also been working on declaratively installing extensions. It doesn't support Raycast extensions yet, as I haven't found a way to fetch just the desired extension from the massive [raycast/extensions](https://github.com/raycast/extensions) repository.

You can check it out here: [kaan-w/vicinae-extensions](https://github.com/kaan-w/vicinae-extensions)

aurelleb:
ah, I think @schromp is working on something similar

schromp:
yeah this is similar to what i have done so far. this will have the same problem with the unwritable support directory though.
havent gotten around to work on that further

Issue #299, opened 2025-09-14
Nix: declaratively install extensions


schromp:
We can install extensions declaratively using nix.

One problem that has to be fixed before this is that extensions need a support directory, which is currently in the same directory as the extension itself.
This doesnt work with nix derivations because they are read only.

Current work on this:
https://github.com/schromp/vicinae/tree/nix-extension-support by @schromp 
https://github.com/kaan-w/vicinae-extensions/tree/main by @kaan-w

aurelleb:
I guess we could store the support directory elsewhere if that's a problem. That's a breaking change however.

schromp:
> I guess we could store the support directory elsewhere if that's a problem. That's a breaking chance however.

I opened a pr which would do that. Maybe you can look it over and see if its acceptable

aurelleb:
we merged this btw

aurelleb:
can we close this @schromp ?

schromp:
not really as there is still work to do in the nix module to make installing extensions possible.
The links above are 2 wips. i wanted to adapt some logic from @kaan-w and then upstream but i didnt have time yet

we can close  this issue and i will open  a WIP PR though if you want

aurelleb:
okay let's do that

PR #301, opened 2025-09-14
split up support and extension path

schromp:
Fixes problems in #299 

I hope this is okay like this. In my testing i did not have any problems beside having to rebuild the data from the support directories

### DIFF
--- a/extension-manager/src/index.ts
+++ b/extension-manager/src/index.ts
@@ -50,8 +50,8 @@ class Vicinae {
 		if (request.payload?.load) {
 			const load = request.payload.load;
 			const sessionId = randomUUID();
-			const supportPath = join(load.extensionPath, "support");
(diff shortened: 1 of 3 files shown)


aurelleb:
there is a conflict with manager.proto

schromp:
> there is a conflict with manager.proto

whoopsies

aurelleb:
I don't see anything wrong with doing this, however it would be great if you could add some logic to copy over the old support directory. If a support directory already exists in the old location, just copy it over.

That way we can make the switch seemless and non breaking. We would remove that logic in a few versions.  What do you think?

schromp:
sure, ill update the pr when i have time

schromp:
With the new logic the folders get migrated as soon as the command is called

aurelleb:
works great, thanks!

PR #306, opened 2025-09-15
fix: activate window on show and try to center it

### LINKED ISSUE #290: Make LauncherWindow appear in a fixed position
**Is your feature request related to a problem? Please describe.**
Currently, when opening Vicinae, the LauncherWindow appears in a seemingly random position decided by the window manager, which can be a bit annoying when using a floating window manager.

**Describe the solution you'd like**
It would be nice if Vicinae forced itself to be launched in the center of the primary desktop (or perhaps a setting for the vertical/horizontal anchor of the window), with maybe an optional setting like "Remember last window position" so that if you move the window, and the option is enabled, the next time LauncherWindow appears it will go to the last remembered position.

### LINKED ISSUE #291: Launcher input not focused on launch
**System information**

- Version: v0.9.6 (0da9c871)
- OS: EndeavourOS - 
- QT Platform: wayland

**Describe the bug**

When launching vicinae, the input field is not automatically focused. In order to be able to search, I first have to click in the input field.

**To Reproduce**

Open the launcher and then type. The input will go to whatever was in focus before.

**Expected behavior**

As soon as the launcher is opened, the input field should have focus.

**Additional context**

I'm using Gnome with PaperWM and the vicinae gnome extension installed. I can however reproduce the issue with Paperwm disabled, so I don't think it's related.

aurelleb:
The centering will only work if the windowing system supports it. On Linux Wayland doesn't, X11 does.
Should fix #291
Fixes #290


aurelien-brabant:
Should fix #291

giodamelio:
I think this (or something else) might have caused a regression for focusing in some configurations.

Using dfd7f68008e05bd6b18b1ee55fcf094ce6d57d78 (after this PR was merged) the auto focus isn't working
If I use 0da9c871e1b697d7551687afd4a2bec01d32d288 (from before the PR) the focus works fine.

I am using a Sway on NixOS with these versions:
sway version 1.11
NixOS unstable built on NixPkgs from `e9e0d35e5f735bf3d1e96815272f46fe7083232c`

I can get you more debug data if it would help.

aurelleb:
It could help if you could provide a video of this happening
I suspect this could be related to the change from 'exclusive' keyboard interactivity to 'on demand' for layer-shell.
It seems a bit buggy on Hyprland as well.

dawsers:
I am seeing this too. I am running scroll, which is also a fork of sway.

When the vicinae window is the only one on the monitor, the input field gets focus, but when there are more windows, the one in focus keeps input focus, and you need to click on the vicinae window to activate its input focus.

It is definitely related to keyboard interactivity. I used to change it to OnDemand for running vicinae in the debugger, so I could use the debugger's key bindings. The current behavior is exactly what I was getting back then.

aurelleb:
Ha! Then I'm not sure what we should do, I guess default back on exclusive and maybe allow on demand to be set in the settings for those that still want it?

dawsers:
If you want, I can have a look to see if there is a way to activate the window regardless of the keyboard mode that also works for sway, because I guess it already works for Hyprland, or you would have noticed it.

aurelleb:
It works on Hyprland most of the time but has the same issue on very specific occasions (if the window below it has some input focused for instance). so it's still not ideal on Hyprland. What I've found is that Hyprland refuses to activate the window when that is the case.

dawsers:
It seems the new layer-shell-qt that was released yesterday has some new functions to request activation, https://github.com/KDE/layer-shell-qt/blob/master/src/qwaylandlayersurface.cpp

The default for `activateOnShow` is `true`.

``` cpp
bool QWaylandLayerSurface::requestActivateOnShow()
{
    if (!m_interface->activateOnShow()) {
        return false;
    }

    if (m_interface->keyboardInteractivity() == Window::KeyboardInteractivityNone) {
        return false;
    }

    if (m_window->window()->property("_q_showWithoutActivating").toBool()) {
        return false;
    }

    return requestActivate();
}
```

We could implement this using Wayland protocol extensions, like "activate", or simply wait for this to be available.

dawsers:
> It works on Hyprland most of the time but has the same issue on very specific occasions (if the window below it has some input focused for instance). so it's still not ideal on Hyprland. What I've found is that Hyprland refuses to activate the window when that is the case.

Wayland doesn't let applications activate themselves. They need to go through a process through the xdg_activation extension protocol, request activation, and the compositor has a final word on allowing that activation. It is a security procedure. `layer-shell-qt` implements that process, and in the new version, it allows more flexibility for clients of the library to request activation regardless of the keyboard mode.

aurelleb:
Yeah I was thinking `activateWindow` in QT was using something like that on the wayland platform already, but most likely not. 
It's great if layer-shell-qt is getting it, I guess we can wait for this and revert to exclusive for the time being? Unless you want it sooner and you can try to implement it.

dawsers:
> Yeah I was thinking `activateWindow` in QT was using something like that on the wayland platform already, but most likely not. It's great if layer-shell-qt is getting it, I guess we can wait for this and revert to exclusive for the time being? Unless you want it sooner and you can try to implement it.

I think that is a good idea. It will be cleaner to use their implementation instead of adding a new protocol just for that purpose.

The version you have in `vicinaehq/layer-shell-qt` already contains the function, so in theory, statically linking would work. However, I haven't been able to make it work yet. I don't know the build system very well, and when compiling it is trying to use the system headers instead of the ones downloaded.

aurelleb:
Yeah I tried to statically link it but I ended up having issues with the QT platform plugins for some reason, I'm going to give it another go. The header thing is easily fixed by explicitly adding the include directory I think.

Edit: no so easy actually, as they generate the full headers at install time

aurelleb:
FYI: I was able to fix the header generation in my fork of layer-shell-qt and statically linking now works and finds the right headers.
And this new version seems to have fixed the issue, but I need to test this a bit more.

dawsers:
That's great, thank you!

aurelleb:
okay so #327 is probably related to statically linking layer-shell-qt.

giodamelio:
Wow. Ya'll figured out a lot while I was asleep :laughing: . It sounds like you have this under control, but let me know if there is any more debug info that would be helpful, and I am happy to help test any PRs too!

PR #329, opened 2025-09-19
Fix a SEGV When No Items Match Typed Text

giodamelio:
First let me say that I haven't programmed in C++ since I did embedded stuff like 10 years ago, so this may be way off base. So feel free to tell me to just pound sand and close this PR! I just stumbled around with GDB until I found the spot then had a LLM fix the function based on the backtrace. So I have no context and the only positive point is that it isn't crashing for me anymore :laughing: .

Using latest master I am getting a segfault when I type enough text that it doesn't match anything and should just go to the fallback options. From what I can tell it is because of a list that is assumed to have items in it but doesn't for some reason.

### DIFF
--- a/vicinae/src/ui/omni-list/omni-list.cpp
+++ b/vicinae/src/ui/omni-list/omni-list.cpp
@@ -824,6 +824,11 @@ OmniList::AbstractVirtualItem *OmniList::itemAt(const QString &id) {
 }
 
 bool OmniList::selectFirst() {
+  if (m_items.empty()) {
+    setSelectedIndex(-1);
+    return false;
+  }
+
   for (int i = 0; i < m_items.size(); ++i) {
     auto item = m_items[i].item;
 


aurelleb:
This is very unlikely to be the issue. As you can see, if the  list is empty we simply do not enter the loop already and we also set the selected index to -1.
Can you attach the gdb backtrace please?
And if possible, a short video showing the crash occur? I haven't had a list-related crash in a very long time.

giodamelio:
Ya that does make sense. Not sure why it stops crashing with it though. I did mess with the the build flags in the flake to get debug symbols, but I am pretty sure it was crashing with those on.

I will open an issue tomorrow with all the details.

aurelleb:
it could be an indicator that there is some memory corruption happening somewhere else, and adding a few instructions just moved to problem away, I honestly have no idea. Nobody reported such faults so that's very weird, looking for your video

PR #338, opened 2025-09-20
fix(extensions): Pass command arguments to extension runtime

koutselakismanos:
I was building an extension and couldn't get my command arguments to show up props.arguments was always empty.

I did some digging and it looks like the C++ backend was forgetting to pack the arguments before sending them over to the Node.js manager.

🍺


aurelleb:
perfect, thanks!

Issue #344, opened 2025-09-21
Network requests to Raycast Store buffer for ~25 seconds on first fetch


koutselakismanos:
When you open the Raycast store extension for the first time, it buffers for ~25 seconds. I did some digging, and it looks like there might be a networking issue with the QNetworkAccessManager. For now I found a workaround by disabling http/2.

koutselakismanos:
I think the issue should remain open until we find why does this issue occur.

aurelleb:
Yeah it really depends actually, it doesn't do it always. But you are right I noticed it too. Maybe not ~25 seconds though.

aurelleb:
fixed by #345

PR #345, opened 2025-09-21
Workaround for long buffering in raycast store

koutselakismanos:
This PR is a workaround for the long buffering ~25 seconds when opening the Raycast Store for the first time.
#344


aurelleb:
lgtm, thanks!

Issue #353, opened 2025-09-23
Add Optional Emacs Keybinding Scheme


falood:
First and foremost, thank you for the excellent work and continued dedication to Vicinae!

## **Is your feature request related to a problem? Please describe.**
I'm a big fan of the project's experience. As an Emacs user, I've also heard from others in the community who would appreciate having Emacs-style keybindings for navigation and editing.

## **Describe the solution you'd like**
   - Introduce a "Keybinding Scheme" option in the settings page. This would preserve the current Default scheme (unchanged) and add a new, optional Emacs scheme.
   - Update navigation keys for the Emacs scheme. For list navigation, the Emacs scheme would use Ctrl+N (for down) and Ctrl+P (for up). The Default scheme would remain as-is.
   - Add Emacs-style editing keys in the search bar (Emacs mode only). This would include common bindings like:
       - Ctrl+B (backward-char), Ctrl+F (forward-char)
       - Ctrl+A (beginning-of-line), Ctrl+E (end-of-line)
       - Ctrl+K (kill-line), Ctrl+U (kill-whole-line)
       - Alt+B (backward-word), Alt+F (forward-word)
       - Alt+Backspace (backward-kill-word), Alt+Delete (kill-word)
   - Note: While this may differ from the philosophy or shortcuts of apps like Raycast, the Default scheme will be completely unaffected. This feature is intended to be strictly  opt-in to avoid disrupting existing users.

  Known or Potential Keybinding Conflicts

  I've identified a few potential conflicts with existing shortcuts:

   - `Ctrl+B`: This is currently used to Open Action Panel. In Emacs, Ctrl+B is for moving the cursor backward ("backward-char").
   - `Ctrl+P`: In the Clipboard History, this is the shortcut for Open Filter. In Emacs, Ctrl+P is for navigating to the previous line ("previous-line").
   - `Ctrl+F`: In Emacs, this is for moving the cursor forward ("forward-char"). However, in many applications, Ctrl+F is the standard shortcut for Find, which could be a consideration.

## **Additional context**
  I would like to understand the developers' perspective on keybindings. Is the goal to support optional schemes like Emacs, or is it preferred to maintain a single, consistent user experience for everyone?

[...]

PR #364, opened 2025-09-26
Emacs key binding

falood:
based on the discuss https://github.com/vicinaehq/vicinae/issues/353

this pr added the emacs keybinding, and use ctrl + o to toggle action panel and use alt + p to open the filter in clipboard view.

I tried to use alt + x for the action panel, but it seems like opt modifier doesn't work well.

BTW: I'm not  familiar with either c++ and qt, most of codes are generated by AI, but I compiled it locally with nix build and it works well.


aurelleb:
The code is actually quite good imo. Eventually I would like the editing stuff to work in all inputs, not only the search bar, so we will have to factorize it. Also we will have to think about how we can implement modality for keybinding schemes that need it (vim).
But it works great as it is, so I'm merging this, thank you very much!

PR #418, opened 2025-10-03
feat: track recently accessed files for all file actions, not just primary ones

dagimg-dot:
Previously, only the primary action (first action in the action panel) would add files to the recently accessed list. Secondary actions like "Open in folder", "Copy file path", and other suggested app openers were not being tracked.

This PR introduces a `FileActionWrapper` that automatically tracks file access for any file-related action, ensuring comprehensive tracking across all interaction methods in the search files view.

**Changes:**
- Added `FileActionWrapper` class to wrap non-primary file actions
- Modified `FileListItemBase` to use the wrapper for utility and secondary actions
- All file actions now call `files->saveAccess()` regardless of how they're accessed


aurelleb:
I'm not sure I'm a big fan of having it tracked for simple things like copying the name or path. For the open app/ in folder action I understand but for the rest I find that a bit weird imo.

dagimg-dot:
> I'm not sure I'm a big fan of having it tracked for simple things like copying the name or path. For the open app/ in folder action I understand but for the rest I find that a bit weird imo.

yeah you are right, i was going to ask you but you were not online at the time. lets remove it for the others.

dagimg-dot:
> I'm not sure I'm a big fan of having it tracked for simple things like copying the name or path. For the open app/ in folder action I understand but for the rest I find that a bit weird imo.

Check it now i simplified it and its as intended.

aurelleb:
thanks!

PR #429, opened 2025-10-04
fix: emacs action toggle key

### LINKED ISSUE #403: `CTRL   O` opens the selected app instead of opening menu
**System information**

- Version: v0.13.4 (aaa98d87)
- OS: CachyOS - 
- QT Platform: wayland

**Describe the bug**
in emacs keybinds selected,
if we press ctrl + o, instead of opeing options menu, it opens the app.

Image

**To Reproduce**

1. select emacs mode keybinds
2. now open vicinae, select any app, then press ctrl+0
3. instead of opening the menu, it will open the app.

**Expected behavior**

IT opens the menu

**Screenshots**

given above.

**Additional context**

N/A

idlip:
- Ctrl+O is also bound to "Open Location" which opens desktop entry in editor which takes precendence
- Re-bound to Ctrl+H and Ctrl+. to act

- Fixed #403

Although this is simple keybinding fix, I've compiled and tested.
C-h and C-. toggle works!

### DIFF
--- a/vicinae/src/ui/launcher-window/launcher-window.cpp
+++ b/vicinae/src/ui/launcher-window/launcher-window.cpp
@@ -221,11 +221,14 @@ bool LauncherWindow::event(QEvent *event) {
   if (event->type() == QEvent::KeyPress) {
     auto keyEvent = static_cast<QKeyEvent *>(event);
 
-    // Toggle action panel: Default Ctrl+B ; Emacs Ctrl+O
-    auto config = ServiceRegistry::instance()->config();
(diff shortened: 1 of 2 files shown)


aurelleb:
I should have opened a draft PR but I'm currently working on making the action shortcut customizable from the settings (as action button customization goes beyond just the input scheme)

idlip:
Ah thats nice.
I'll close this then.

Please take your time we can wait... Hopefully if possible you can make all bindings customizable via config ? :-)

aurelleb:
yeah eventually most will be made configurable. Ideally we should also be able to create global keybinds from vicinae directly.

PR #436, opened 2025-10-06
Fix instructions to toggle Vicinae window after starting the server

Kricheldorf:
Hi. I'm not sure it was a regression from the latest changes. If it was, please feel free to close this PR.

After updating, I realized the `vicinae` command was not toggling the window anymore but the server still presented the following message:

image

This PR fixes the message by mentioning the `vicinae toggle` command to toggle the window instead of the `vicinae` command without arguments.

For reference, this is the behavior of the `vicinae` command:

image

(it only displays information about the CLI).

Sorry for creating a PR directly, let me know if I should have created an issue first (I didn't find any existing issue about this problem). Thanks for the great work, it's being a pleasure daily driving Vicinae here :raised_hands:

### DIFF
--- a/vicinae/src/cli/server.cpp
+++ b/vicinae/src/cli/server.cpp
@@ -229,7 +229,7 @@ void CliServerCommand::run(CLI::App *app) {
 
   ctx.navigation->launch(std::make_shared<RootCommand>());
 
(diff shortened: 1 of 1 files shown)


aurelleb:
you're right, I simply forgot to update this one! Thanks!

PR #465, opened 2025-10-10
fix: dont use nixos-unstable in cachix workflow

schromp:
The nixpath was only used for nix-shell which i migrated to the never cli version so it uses the flake input.
This makes us not depend on nixos-unstable anymore because we really dont need that anywhere else.


aurelleb:
it passes!

PR #474, opened 2025-10-11
feat: add Zendesk icon

cluther:
For shortcuts like https://mycompany.zendesk.com/agent/tickets/{argument}.


cluther:
I think I got all of the places that need to be updated. Sorry for the large set of changes. The `build-icon-qrc.js` script did a lot of reformatting.

aurelleb:
that's just fine thank you!

PR #492, opened 2025-10-15
chore: add compile_commands.json to .gitignore

kfatyuip:
The compile_commands.json file is automatically generated by build systems such as CMake using commands like:

```bash
cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ...
```

This file is primarily used for code analysis tools (e.g., clangd) and local development setups, such as Neovim with LSP plugins. It contains machine-specific compilation information and should not be tracked by git. Adding it to .gitignore helps prevent accidental commits and keeps the repository clean.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ extension-manager/runtime/node_modules
 /vgcore*
 .env
 .direnv
+compile_commands.json
 
 vicinae/assets/extension-runtime.js
 


aurelleb:
the compile_commands.json file is placed into the build directory which is itself ignored, how did you manage to have this issue?

kfatyuip:
Some extensions of vscode like makefile tools or toolchains caused compile_commands.json to be generated in the root to be searched at a higher priority for indexing, but standard process would not lead to this. Sorry for worrying too much.

PR #493, opened 2025-10-16
vicinae.desktop: add %u for deeplinks to work

rithvikvibhu:
Looks like deeplinks require %u in the Exec to work:

```
❯ xdg-open vicinae://toggle
kf.kio.core: "/usr/share/applications/vicinae.desktop" contains supported protocols but doesn't use %u or %U in its Exec line! This is inconsistent.
```

and

image

System: Manjaro rolling / Wayland / KDE


aurelleb:
yes that's correct. If only all desktop files could do it properly, we would have way less problems!

PR #503, opened 2025-10-19
[GNOME] Paste to Active window implementation

dagimg-dot:
## Summary
Implements reliable "paste to active app" functionality for GNOME environment by fixing clipboard synchronization issues.

## Problem
- Clipboard setting was asynchronous, causing paste shortcuts to execute before clipboard content was updated

## Files Changed
- `clipboard-server.hpp/cpp`: Added virtual clipboard setting method
- `gnome-clipboard-server.hpp/cpp`: Implemented GNOME-specific synchronous clipboard setting
- `clipboard-service.cpp`: Fixed copyText() to use proper clipboard server path (essential for emoji view paste action)

## Solution
- Added `setClipboardContent()` virtual method to `AbstractClipboardServer` for DE-specific clipboard handling
- Implemented synchronous clipboard setting in `GnomeClipboardServer` using D-Bus calls to GNOME extension
- Fixed `copyText()` to use `copyQMimeData()` path instead of direct Qt clipboard calls
- Added comprehensive MIME type support via `SetContentBinary` D-Bus method
- Other DEs continue to use the default Qt clipboard


aurelleb:
lgtm, thanks!

PR #507, opened 2025-10-20
fix: use the correct package in home-manager module

schromp:
Im not sure i can really test this outside of this repo because the derivations wouldnt match then (atleast i didnt find a way).

This _should_ fix having to specifiy `package = inputs.vicinae.packages.${pkgs.system}.default;` in order to get the cachix hit


aurelleb:
thanks!

Issue #22, opened 2025-08-15
Action panel not showing up on keyboard shortcut or click


aurelien-brabant:
> https://github.com/user-attachments/assets/c7d69cf8-7ac6-47d6-8a3a-f4776e1f0a99
> 
> I also noticed the same issue, but the cmd + B hotkeys take the focus but doesn't display the options menu.
> 
> WM : Niri
> OS : Endeavour OS ( 6.15.9-arch1-1)
>  

 _Originally posted by @C4n4m1 in [#20](https://github.com/vicinaehq/vicinae/issues/20#issuecomment-3191371183)_

I noticed this as well but this is very inconsistent, I didn't find a way to reliably reproduce.

uiriansan:
Same behavior, Niri as well:

https://github.com/user-attachments/assets/99d9a62c-63fd-4b88-9a7f-ac6e797336dd

aurelleb:
Did you do something special to replicate this? or is this seemingly random? Also does it happen if you try right after the server is started?

uiriansan:
> Did you do something special to replicate this

I don't think so.

> Also does it happen if you try right after the server is started?

Yes. 

Maybe some Niri config/rule is conflicting. I will play with them later and take a look at the code as well. I remember having issues in the past with both GTK and QML's built-in popup widgets, so maybe it's a similar issue.

aurelleb:
> > Did you do something special to replicate this
> 
> I don't think so.
> 
> > Also does it happen if you try right after the server is started?
> 
> Yes.
> 
> Maybe some Niri config/rule is conflicting. I will play with them later and take a look at the code as well. I remember having issues in the past with both GTK and QML's built-in popup widgets, so maybe it's a similar issue.

Well, the thing is that the action panel and other absolutely positioned surfaces (dialogs) are *not* popup widgets because it didn't work with wlr-layer-shell (it worked okay when I was still dealing with a floating window, although I had other problems). They are just simple widgets. Maybe that has to do with the issue. Given the fact that it's not easily replicable on Hyprland I think  I might be doing something slightly "undefined" somewhere.

aurelleb:
I think I detected a pattern, it seems to happen a lot more frequently when the launcher changes monitors.

uiriansan:
I couldn't find anything wrong with my Niri config, and I'm afraid I'm way too dumb to read C++ code. I will keep looking, tho.

[...]

PR #525, opened 2025-10-22
feat(script): add installation script with additional update functionality

dagimg-dot:
- add a new `install.sh` for easy installation of vicinae using one command. 
- It uses the appimage bundle, extracts it at `~/.local/vicinae` and  symlink the main vicinae binary to `~/.local/bin`
- you can add the new installation command to the readme. I didn't do it because it might mislead people before this pr is merged.
- I also opened pr [#22](https://github.com/vicinaehq/docs/pull/22) on the docs site to expose this script at https://vicinae.com/install.sh

you can test it using

```
curl -fsSl https://raw.githubusercontent.com/dagimg-dot/vicinae/refs/heads/feat/install-script/scripts/install.sh | bash
```

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -327,7 +327,22 @@ main() {
 
 		echo ""
 		echo "🎉 Vicinae $latest_version has been successfully installed!"
-		echo "You can now run 'vicinae' from anywhere in your terminal."
+		echo ""
+
+		local local_bin_path="$HOME/.local/bin"
+		if [[ ":$PATH:" == *":$local_bin_path:"* ]]; then
+			echo "✓ ~/.local/bin is already in your PATH"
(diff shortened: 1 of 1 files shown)


aurelleb:
it's a very nice script although I think we should also install/symlink the themes and the desktop files as the regular packages do. But currently the appimage doesn't bundle them, I'm going to fix this.
I'm merging this like that and I will add the missing logic for these things myself

dagimg-dot:
> it's a very nice script although I think we should also install/symlink the themes and the desktop files as the regular packages do. But currently the appimage doesn't bundle them, I'm going to fix this.
> I'm merging this like that and I will add the missing logic for these things myself

I was going to change this to draft and do that but I guess it would be better if u do it since the appimage doesn't bundle them . 👍

PR #555, opened 2025-10-26
fix: Improved matugen theme

Sridou:
The current hover selection was more on the grey-scale, and combined with lower opacity, made it a poor contrast ratio. This primarily fixes that, combined with other minor improvements


aurelleb:
thanks!

PR #559, opened 2025-10-27
feat: fanotify filesystem watch (last)

### LINKED ISSUE #538: Support for fanotify's FAN_MARK_FILESYSTEM
From the [doc](https://docs.vicinae.com/file-search#watcher-experimental) it would seem that each folder is watched individually, which from the doc means that a limit can be reached (and I would assume is not as efficient):
```
You can add paths to 'Watcher paths' in the settings to watch for real-time changes with [e-dant/watcher](https://github.com/e-dant/watcher).

There are a number of caveats:

There is no way to watch 'unwanted' directories like node_modules and hidden paths
It requires an inotify instance for every subdirectory, and you can hit the system limit. You can increase this limit by:
```

From the official [man](https://www.man7.org/linux/man-pages/man2/fanotify_mark.2.html) page for fanotify, it would seem it is possible to mark a folder which will in turn watch all contained files and directories:
```
       FAN_MARK_FILESYSTEM (since Linux 4.20)
              Mark the filesystem specified by path.  The filesystem
              containing path will be marked.  All the contained files
              and directories of the filesystem from any mount point will
              be monitored.  Use of this flag requires the CAP_SYS_ADMIN
              capability.
```

In a situation where one wants to index whole disks, this would be an interesting solution.
I would be willing to take on the work if it is something that would interest you.
I have not analyzed how exclusions would fall within this process though.

AntoineGS:
This is still a work in progress and contains added logging/output and code to refactor but it works to give you an idea of what it could be like.

Fixes #538

Todos
- [ ] Cleanup logging
- [ ] Add mentions for BTRFS in the doc or UI, possibly a warning when detected

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,3 +95,7 @@ For more details on installation, usage, configuration, extension development, a
     alt="extension settings"
   />
 </p>
+
+# TEMP NOTES
+
+sudo setcap cap_dac_read_search,cap_sys_admin=+ep ./vicinae
--- a/tempnotes.md
+++ b/tempnotes.md
@@ -1,0 +1,9 @@
+# TEMP NOTES
+
+sudo setcap cap_dac_read_search,cap_sys_admin=+ep ./vicinae
+
+## TODO
+
+fs:
+Since it marks the filesystem, ensure that it only uses mark_filesystem_mount when the folder is a mount point?
+note/warning about bftrs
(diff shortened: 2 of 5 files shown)


quadratech188:
This looks more like an [e-dant/watcher](https://github.com/e-dant/watcher) (the external library that provides `watcher.hpp`) pull request rather than a vicinae pull request, so maybe we should consider contributing to that instead?

quadratech188:
Technically we could have our own version, but then updating the library would be a nightmare

AntoineGS:
😲 I had no idea this was an external dep, my bad.
Yeah you are right, I will see about upstreaming the changes and report back on how that goes.

aurelleb:
I think it's okay to make local edits to our `watcher.hpp` to suit our use cases, and propose a PR upstream if they are interested.

AntoineGS:
I am starting to get conflicts in between branches (I try to keep features/fixes separate as much as possible).
Do you have any objections that I branch each PR off another PR and I number them in the order that they should be merged to keep changes separate?

PR #560, opened 2025-10-27
fix: prevent infinite loop on file-indexer.db when watched (1)

AntoineGS:
When the database file is watched, inotify goes on an infinite loop of getting a database file change event, saving it to the database, which triggers a file change event, etc.

This prevents the database file from being processed.
Feel free to improve the logic, as to me it does not look ideal to have the filename hardcoded but I could not find another way to do it with my (still) limited knowledge of the codebase!

### DIFF
--- a/vicinae/src/services/files-service/file-indexer/file-indexer.cpp
+++ b/vicinae/src/services/files-service/file-indexer/file-indexer.cpp
@@ -65,7 +65,8 @@ void FileIndexer::start() {
   }
 
   for (const auto &entrypoint : m_watcherPaths) {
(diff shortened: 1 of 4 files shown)


aurelleb:
I think we should not hardcode this edge case in the watcher class itself but add an API to the watcher so that we can specify a list of patterns to ignore, to keep the watcher generic enough
@quadratech188 what do you think ?

AntoineGS:
I recently found out that this is a third party library 🤣 so yeah hardcoding it there makes even less sense.

AntoineGS:
Ok I think this is much cleaner now, plus it supports a potential future change in the database name/path!

PR #563, opened 2025-10-28
fix CopyToClipboard in extension error view

rithvikvibhu:
Copy to clipboard wasn't copying the error text, this reads the error.


aurelleb:
thanks!

PR #570, opened 2025-10-29
chore: add clangd to ignore -mno-direct-extern-access

AntoineGS:
This is only to remove the warning clangd puts at the top of every file in this project.


aurelleb:
Ah! thanks for figuring this out, I was too lazy to look it up :rofl:

PR #579, opened 2025-10-30
Form.Description and Form.Separator

rithvikvibhu:
# Description

Descriptions are now displayed:

extension: `word-count` (the counts, etc above the text area is a description)

image

Description with title:

extension: `video-downloader`

image

# Separator

extension: `cobalt` (below url field)

image


aurelleb:
really good one, thanks! The form description is technically not a `JsonFormWidget` but we will fix that later.

PR #594, opened 2025-10-31
Fix qalculate backend

leiserfg:
Without this change I always get `0` as output, seems like qalculate does not like the `=`


leiserfg:
I just figured out that vicinae does not use `=` to trigger calc mode like other launchers :man_facepalming:

PR #600, opened 2025-10-31
feat: improvements to incremental search (2)

AntoineGS:
Instead of limiting ourselves to a depth of 5, we can traverse the whole tree and use the mtime to decide which ones to process.

Due to the changes done for this, it incidentally fixes the issue resolved by https://github.com/vicinaehq/vicinae/pull/586 so I have closed that one:
>If an increment scan were to fail once, incremental scans would be disabled indefinitely because they are not cleaned up, and thus stay in the startedScans list.
>This ensures that increment scans are scheduled at launch if a full scan is not requested.

### DIFF
--- a/vicinae/src/services/files-service/file-indexer/file-indexer-db.cpp
+++ b/vicinae/src/services/files-service/file-indexer/file-indexer-db.cpp
@@ -149,12 +149,47 @@ FileIndexerDatabase::ScanRecord FileIndexerDatabase::mapScan(const QSqlQuery &qu
   return record;
 }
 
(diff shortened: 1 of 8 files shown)


aurelleb (commented):
there are conflicts since I merged the (1)

AntoineGS:
Weird when I rebased it there were not errors, but in any case I saw your first question and realized I had left traces of my initial approach using the last modified timestamp so I cleaned it up!

AntoineGS:
No pressure but in case you were waiting on me the PR is ready to be merged on my end!

aurelleb:
> No pressure but in case you were waiting on me the PR is ready to be merged on my end!

yep, it's all good, thanks!

PR #602, opened 2025-10-31
form: add LinkAccessory

rithvikvibhu:
https://developers.raycast.com/api-reference/user-interface/form#form.linkaccessory

extension: `video-downloader` has a link search accessory ("Supported Sites" below)

image

---

I wanted to uncomment its use in the boilerplate extension, but that is on the latest release of `@vicinae/api` so won't work as-is. Maybe that can be linked to dist of the build outside? idk.


aurelleb:
for the extension boilerplate, since it's not meant to be used directly by end users we can probably link it locally

rithvikvibhu:
Rebased and linked to the api dir.

PR #620, opened 2025-11-02
add Action.CreateQuicklink

rithvikvibhu:
extension: `random`
image
image

extension: `notion`
image

extension: `devdocs`
- author and extension name are undefined, but this PR is only for creating quicklinks
- extension source detection (required for shortcuts to deeplinks) + environment.* values coming in the next PR
image

### DIFF
--- a/vicinae/src/extension/extension-navigation-controller.hpp
+++ b/vicinae/src/extension/extension-navigation-controller.hpp
@@ -33,7 +33,14 @@ public:
 
   std::vector<ExtensionViewWrapper *> views() const { return m_views; }
 
-  void handleViewPoped() {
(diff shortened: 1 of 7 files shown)


vicinae/src/extension/extension-navigation-controller.hpp
```diff
@@ -33,7 +33,14 @@ class ExtensionNavigationController : public QObject {
 
   std::vector<ExtensionViewWrapper *> views() const { return m_views; }
 
-  void handleViewPoped() {
+  void handleViewPoped(const BaseView *view) {
```
aurelleb: it's fine for now but we will have to think about something else, because that's definitely not pretty

aurelleb (approved):

PR #633, opened 2025-11-03
form: add Form.DateTime

rithvikvibhu:
Adds date and datetime picker. Fields added to boilerplate form demo.

The input field looks different for some reason:
image

There is a calendar date picker, but I couldn't find any Qt gui for time. So even datetime shows only a calendar (and time has to be changed in text).

image

It doesn't look great, but makes a few extensions not crash:

extension: `raycast-timezone-converter`
image

fyi: there is a bug where on the page above, submitting when the dropdown is selected crashes vicinae. But submitting with datetime focus works normally. Will take a look at it later.

### DIFF
--- a/extra/extension-boilerplate/src/form.tsx
+++ b/extra/extension-boilerplate/src/form.tsx
@@ -40,6 +40,14 @@ export default function FormElements() {
 			<Form.DatePicker
 				id="dateOfBirth"
 				title="DatePicker"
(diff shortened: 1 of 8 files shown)


aurelleb:
yeah, we probably need to apply a style sheet for the QDateTimeEdit stuff, which we derive from the current vicinae theme. It is what's done for some of the input customization already.

PR #636, opened 2025-11-04
form: add FilePicker and support all props

rithvikvibhu:
# Add <Form.FilePicker />

boilerplate form elements:
image

# Button text based on options

See the boilerplate form example above ^ (pick a file, pick a dir, etc.)

# Fix single directory select

Before, m_multiple was always true, and extensions wouldn't like it:
```cpp
void FilePicker::setMultiple(bool value) { m_multiple = true; }
```

extension: `downloads-manager` - has an option to set a custom downloads folder, but since the picker was always multiple, it returned a list of strings instead of a single string.
image

extension: `cobalt`
image

### DIFF
--- a/extra/extension-boilerplate/src/form.tsx
+++ b/extra/extension-boilerplate/src/form.tsx
@@ -1,4 +1,4 @@
-import { Action, ActionPanel, Form } from "@vicinae/api";
+import { Action, ActionPanel, Form, showToast, Toast } from "@vicinae/api";
 
 // Examples taken from:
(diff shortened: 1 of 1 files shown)


aurelleb:
ah it looks like there are conflicts now.
On a side note, something else is needed for layer shell setups in order to properly support file picking. Right now the file picking dialog is opened below the window. 
Which is funny because you are on KDE, don't you have that issue?

rithvikvibhu:
Rebased btw

PR #638, opened 2025-11-04
Render last top level element instead of first

rithvikvibhu:
If extensions have more than one top level element, raycast renders the last one.

https://github.com/raycast/extensions/blob/main/extensions/google-meet/src/manual-profile.tsx

Google Meet has a fragment like:
```tsx
<>
  <List />
  {condition && <Form />}
</>
```

And when condition is true, raycast displays the form (and hides the list)

now:
image


aurelleb:
maybe we should handle this at the react reconcilier level, as it doesn't make sense to have them all sent through IPC

rithvikvibhu:
moved to reconciler.

rithvikvibhu:
this seems to be breaking timezone converter's Custom Time page.

PR #649, opened 2025-11-05
feat: add deeplink support for opening shortcuts by name

carmilso:
This enables better integration with external applications and automated workflows by providing a user-friendly way to open shortcuts programmatically.

- Add deeplink support that allows opening specific shortcuts via URL using their name instead of ID
- Implement case-insensitive name lookup in ShortcutService with proper error handling and user feedback

Example of usage for Hyprland:
```
bind = $mainMod, G, exec, vicinae 'vicinae://extensions/vicinae/shortcut/open?arguments={"name":"Google search"}'
```

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -5,6 +5,6 @@
 # -DVICINAE_GIT_COMMIT_HASH=XXX (short rev)
 
 release:
-  tag: "v0.16.1"
-  rev: "478a9ee18751ca4f64cd5eb145f206a8601e7d87"
-  short_rev: "478a9ee18"
+  tag: "v0.16.2"
+  rev: "b99015bc2870f9ad3139fab438ac68e3a0bbec0f"
+  short_rev: "b99015bc2"
--- a/vicinae/src/extend/list-model.cpp
+++ b/vicinae/src/extend/list-model.cpp
@@ -25,10 +25,10 @@ ListItemViewModel ListModelParser::parseListItem(const QJsonObject &instance, si
   auto children = instance.value("children").toArray();
 
   model.id = props["id"].toString(QString::number(index));
-
   model.title = props["title"].toString();
   model.subtitle = props["subtitle"].toString();
-  model.icon = parseListItemIcon(props.value("icon"));
+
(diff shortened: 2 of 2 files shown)


aurelleb:
while this works, I'm not sure implementing this as a command is the best way to do it, as a shortcut is not really an extension properly speaking.
I was thinking about handling it as a separate case here: https://github.com/vicinaehq/vicinae/blob/de12226d691f084bd2033239be8243a11451f3c3/vicinae/src/ipc-command-handler.cpp#L170

aurelleb:
and then something like that: `vicinae://shortcuts/<shortcut_id>?arguments=blahblahblah`

carmilso:
Thanks for the suggestion, that makes more sense. I rewrote the commits and implemented it your way. Now you can do: 

- vicinae "vicinae://shortcuts/Google search"
- vicinae vicinae://shortcuts/Google%20search
- xdg-open "vicinae://shortcuts/Google search"

carmilso:
Only thing I kept is the name instead of the ID, because the user knows the name, but not the ID that vicinae assigned to that shortcut.

rithvikvibhu:
Are names unique? Can 2 shortcuts have the same name?

aurelleb:
@rithvikvibhu they aren't.

@carmilso we need to use the ID here as well, this is not an issue. Like we do for extensions, we can add a 'copy deeplink' action to make it easier to get the full deeplink

luigieli:
Sorry to ask, but is there any news on this? I'd love to have this feature.

aurelleb:
@luigieli yes it just shipped in v0.20.8! You can copy the deeplink from the entry in the root search directly now.
Also I'm closing this. Full documentation is at https://docs.vicinae.com/deeplinks

luigieli:
> @luigieli yes it just shipped in v0.20.8! You can copy the deeplink from the entry in the root search directly now. Also I'm closing this. Full documentation is at https://docs.vicinae.com/deeplinks

Wow, thanks @aurelleb! The timing was perfect.

PR #675, opened 2025-11-09
fix (nix module): use TOML parser instead of JSON for themes

Aceroph:
Untested, although I am confident it will work.

### DIFF
--- a/extra/extension-boilerplate/.gitignore
+++ b/extra/extension-boilerplate/.gitignore
@@ -1,1 +1,2 @@
 /node_modules
+vicinae-env.d.ts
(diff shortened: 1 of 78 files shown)


aurelleb:
needs testing

Aceroph:
> needs testing

Now tested, the file is generated using the same format as before with the exception of using the toml file extension. Now I am testing my [second change](https://github.com/vicinaehq/vicinae/pull/675/commits/3815e305eb5df989e3d700bebbab12866270e721) of moving the generated theme files to XDG_DATA_HOME, since I believe Vicinae is now looking there.

Aceroph:
My custom theme is now applied to Vicinae, I think my PR is ready for review @aurelleb

aurelleb:
thanks!

PR #682, opened 2025-11-09
feat: Consider IME preedit strings as search text

quadratech188:
Input Method Editors work by taking control of the last $N$ characters of user input, and adding additional logic that manipulate these characters before sending them off to the program. What this means however is that Vicinae (by default) doesn't know about these characters until later, and it feels like the search results 'lag behind' the input.

This PR makes Vicinae consider the temporary IME text as part of the search text.

Note that **this has only been tested with Korean** (More specifically [kime](https://github.com/Riey/kime)), and **I have no idea if it works or is even a good idea with other languages**. (I should probably add a toggle)

(See first commit's description for why the refactor is here)

### DIFF
--- a/vicinae/src/navigation-controller.cpp
+++ b/vicinae/src/navigation-controller.cpp
@@ -56,7 +56,7 @@ void NavigationController::broadcastSearchText(const QString &text, const BaseVi
 
 void NavigationController::setSearchText(const QString &text, const BaseView *caller) {
   broadcastSearchText(text, caller);
-  if (caller == topView()) {emit searchTextTampered(text);}
+  if (caller == nullptr || caller == topView()) {emit searchTextTampered(text);}
 }
 
 void NavigationController::setSearchPlaceholderText(const QString &text, const BaseView *caller) {


quadratech188:
I'm having some trouble making this behavior optional.
I currently hijack `Searchbar::textEdited` to also send text with IME events, but I can't access Vicinae settings from `SearchBar`.
The only option seems to be creating a separate signal chain for IME events through `SearchBar`, `GlobalHeader`, and `NavigationController`, but this seems a bit messy. It also puts the concern of composing the IME events on `NavigationController`, which is really not what it's for.

aurelleb:
you can access it from anywhere using the service registry, although that's not pretty:  https://github.com/vicinaehq/vicinae/blob/5109e081d28458f4b233a234204624ec66435165/vicinae/src/ui/search-bar/search-bar.cpp#L66 (this a bad example)

But I guess the cleaner way to do it would be to add a method to configure an option on the SearchBar and call that method from the GlobalHeader, using `context->services` directly instead of using the singleton.

quadratech188:
> you can access it from anywhere using the service registry, although that's not pretty: 

How did I fail to see that bit, it was not even 60 lines away from where I was writing code!

I don't see any config-related code in GlobalHeader(It's all in NavigationController) and checking config directly seems simple enough, so I'll just do that.

Thanks for the fast response!

aurelleb:
I mistook this for another call, don't worry about it

aurelleb:
I tried it out, it doesn't seem to cause an particular issue with my Japanese IME (mozc, fcitx). Also doesn't seem to conflict with "normal" search using latin text so I don't know if having a toggle is all that useful. I guess it makes sense to have it be opt-in until we can maybe find additional quirks

quadratech188:
Really? Because I just found a bug that always inserts the preedit string in the front.

aurelleb:
@quadratech188 hmmm didn't catch that. but I have a very limited dataset to search on so idk

PR #689, opened 2025-11-11
nix(home-module): refactor & improve home-module

mochouaaaaa:
Clean up the code, improve implementation robustness, and align with standard practices. This PR contains no new features or breaking changes.

-  Self-reviewed my code
-  No new warnings or errors

I replaced builtins.toJSON with the type and generate functions provided by pkgs.formats.json, which is the standard practice in Home Manager.

- befer:
The json format is not maintained

- after
image

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -1,4 +1,5 @@
-self: {
+self:
+{
   config,
   pkgs,
   lib,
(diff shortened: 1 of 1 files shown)


aurelleb:
thanks!

leiserfg:
Guys, do you think it makes sense to keep the home module here instead of using the one from Home-Manager?

Issue #683, opened 2025-11-09
bug(List.Item.Detail.Metadata.Label): `text` property does not support color `{ color?: Color; value: string }`


knoopx:
**System information**

- Version: v0.16.2 (unknown)
- Build info: GCC 14.3.0 - Release
- Provenance: nix
- OS: NixOS 25.11 (Xantusia) - 25.11
- QT Platform: wayland
- DE: niri/Wayland

**Describe the bug**

`List.Item.Detail.Metadata.Label` `text` property does not support color notation`{ color?: Color; value: string }`, vicinae renders nothing instead

**To Reproduce**

```javascript
<List.Item.Detail.Metadata.Label
   text={{
    value: "Hello",
    color: "#ff000"
}} />
```

**Expected behavior**

renders label with color

https://developers.raycast.com/api-reference/user-interface/list#list.item.detail.metadata.label

PR #715, opened 2025-11-13
feat(ColorLike): List.Item.Detail.Metadata.Label#text support for { color?: Color; value: string }

knoopx:
fixes https://github.com/vicinaehq/vicinae/issues/683


aurelleb:
thanks!

PR #723, opened 2025-11-15
nix: create config file only when configured

### LINKED ISSUE #719: Nix home-manager module creates config files regardless of input
**System information**

- Version: v0.16.5 (576ea047d)
- Build info: GCC 14.3.0 - Release
- Provenance: nix
- OS: NixOS 25.11 (Xantusia) - 25.11
- QT Platform: wayland
- DE: niri/Wayland

**Describe the bug**

#689 changed the behavior of home-manager module. Earlier `.config/vicinae/vicinae.json` was created only when `programs.vicinae.settings` was filled by the user. This PR stripped that logic, so the file is created regardless. This is a problem for users who don't want to use the module for configuring vicinae and instead use normal methods, because home-manager fails to build a new generation complaining that the old file will be clobbered.

Vortriz:
Closes #719. Please read that for more information.

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -108,7 +108,7 @@ in
           ]);
         in
         valueType;
-      default = { };
+      default = null;
       description = "Settings written as JSON to `~/.config/vicinae/vicinae.json.";
       example = lib.literalExpression ''
         {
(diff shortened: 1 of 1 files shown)


aurelleb:
thanks!

PR #726, opened 2025-11-15
fix(extension): support List.Item title and subtitle props alternate usage with tooltip.

Bizangel:
Currently the raycast api supports tooltip usage using `{ tooltip?: string; value: string }` on some List.Item properties such as title, subtitle and icon.

Icon is already handled as part of https://github.com/vicinaehq/vicinae/commit/b466d44b3efeea6031ff0f9b023ce42097dd86fd

However title and subtitle are missing. If we use the tooltip form - no text is displayed as the `.toString()` returns empty string from an object.

We make it use the `value` property correctly.

Ref: https://developers.raycast.com/api-reference/user-interface/list#list.item


aurelleb:
nice catch! thank you!

PR #733, opened 2025-11-17
Add COPR packaging and automation for vicinae

dacrab:
## Summary

This PR adds COPR packaging support and build automation for vicinae so that Fedora users can enable a COPR repo and install vicinae via the package manager.

Key pieces:
- A Fedora RPM spec file for vicinae
- A small helper script to build an SRPM and trigger a COPR build via `copr-cli`
- A GitHub Actions workflow that builds the SRPM and triggers a COPR build on tagged releases
- Minor ignore/Makefile tweaks so the local RPM workflow is smooth

## Motivation

I maintain a COPR project for vicinae and wanted an official, reproducible way to:

1. Build vicinae as an RPM using upstream sources.
2. Keep the COPR repo up to date from upstream changes.
3. Let Fedora users install vicinae by enabling `copr enable killcrb/vicinae` and running `dnf install vicinae`.

These changes live entirely in packaging/CI and do not affect the runtime behavior of vicinae itself.

### 1. New RPM spec: `vicinae.spec`

A Fedora-style spec file that:

- Uses the upstream GitHub tag tarball as `Source0`:
  - `Source0:  %{url}/archive/refs/tags/v%{version}.tar.gz`
- Uses GitHub’s cmark-gfm fork from COPR as a system dependency instead of building it from source:
  - `BuildRequires: cmark-gfm-devel`
  - `Requires:      cmark-gfm`
- Configures the CMake build similarly to your existing Release configuration:
  - `-DCMAKE_BUILD_TYPE=Release`
  - `-DTYPESCRIPT_EXTENSIONS=ON`
  - `-DINSTALL_NODE_MODULES=ON`
  - `-DWAYLAND_LAYER_SHELL=OFF`
  - `-DUSE_SYSTEM_CMARK_GFM=ON`
- Declares the other build requirements that are already expected in the project:
  - Qt6 (qt6-qtbase-devel, qt6-qttools-devel, qt6-qtsvg-devel, qt6-qtwayland-devel)
  - protobuf-devel, abseil-cpp-devel, minizip-compat-devel
  - qtkeychain-qt6-devel, layer-shell-qt-devel, wayland-devel
  - nodejs, npm, extra-cmake-modules, git
- Installs and packages the key runtime artifacts:
  - Binary: `%{_bindir}/vicinae`
  - Desktop entries: `%{_datadir}/applications/vicinae.desktop`, `%{_datadir}/applications/vicinae-url-handler.desktop`
  - Icon: `%{_datadir}/icons/hicolor/512x512/apps/vicinae.png`
  - Systemd user unit: `%{_prefix}/lib/systemd/user/vicinae.service`
  - Themes: `%{_datadir}/vicinae/themes`

This spec reflects the current CMake install layout and passes in the CMake options that are already in the tree, so behavior should match your existing builds.

### 2. cmark-gfm FetchContent patch: `cmark-gfm-no-git.patch`

To make the project more friendly to offline or policy-restricted build environments (like RPM build chroots), this patch changes the `FetchContent_Declare` for `cmark-gfm` to use a local tarball URL instead of a Git repository:

- Before (from upstream):
  - `GIT_REPOSITORY https://github.com/github/cmark-gfm`
  - `GIT_TAG        0.29.0.gfm.13`
- After (for packaging / non-system builds):
  - `URL ${CMAKE_SOURCE_DIR}/cmark-gfm-0.29.0.gfm.13.tar.gz`

In the RPM spec, we rely on the system `cmark-gfm` (`USE_SYSTEM_CMARK_GFM=ON`), so this patch is primarily useful for environments that still want to vendor cmark-gfm but cannot do network fetches during the build.

If you prefer not to ship this patch in-tree, it can instead live purely as a packaging artifact; I included it here so the spec and patch are versioned together.

### 3. COPR helper script: `scripts/copr-build.sh`

A small script to make maintaining the COPR project easy from this repo:

- Fetches remote Sources according to `vicinae.spec` using `spectool -g -R`.
- Builds an SRPM into `.srpm/` using `rpmbuild -bs` with the correct `_sourcedir`, `_specdir`, and `_srcrpmdir` definitions.
- Finds the most recent `vicinae-*.src.rpm` and calls:
  - `copr build vicinae <latest_srpm>`

This is wired into the `Makefile` as a `copr-build` target so a maintainer can do:

```bash
make copr-build
```

…to produce and submit a COPR build from whatever commit/tag they have checked out.

### 4. GitHub Actions workflow: `.github/workflows/copr-release.yml`

Adds a workflow that:

- Triggers on:
  - Tag pushes matching `v*` (e.g. `v0.16.6`).
  - Manual `workflow_dispatch`.
- Runs on your existing self-hosted Linux runner labels (`[self-hosted, Linux, X64, prod]`).
- Steps:
  1. Check out the repo with full history.
  2. Ensure RPM tooling and `copr-cli` are installed (`dnf install rpm-build rpmdevtools redhat-rpm-config copr-cli fedpkg spectool`).
  3. Build an SRPM from `vicinae.spec` into `.srpm/` (same process as the local helper script).
  4. Submit the SRPM to the `vicinae` COPR project using `copr build`.

This gives you a single, repeatable path to update the COPR repo whenever you cut a new tag, without having to build or upload SRPMs manually.

### 5. Minor repo hygiene

- `.gitignore` now ignores:
  - `/.build`
  - `/.rpm`
  - `/.srpm`

This avoids committing local rpmbuild artifacts and temporary build trees.

- `Makefile` now has a `copr-build` target that just invokes the helper script.

## COPR project configuration (for context)

Outside this repo, the COPR project `vicinae` was configured to:

- Enable network for builds (`--enable-net on`) so `npm install` in the TypeScript subprojects can reach the npm registry.
- Add the `quadratech188/cmark-gfm` COPR as a build-time repo and a runtime repo dependency so `cmark-gfm` and `cmark-gfm-devel` are available in the Fedora 43 chroot and for users.

These configuration changes are not part of this PR, but the spec and workflow assume that `cmark-gfm` is available from the configured repos.

## Impact

- No changes to the core vicinae codebase behavior.
- Adds first-class RPM packaging support and an automated COPR publishing flow.
- Makes it easy for Fedora users to install vicinae via COPR and for maintainers to keep that repository up-to-date.

## Questions / Notes

- If you’d prefer the spec and COPR workflow to live under a different path (e.g. `packaging/fedora/`), I can move them.
- If you’d rather keep the cmark-gfm patch out of the main tree, I can drop it from the repo and adjust the spec to apply it from the SRPM only.
- Happy to tweak the COPR project name or runner labels as needed to match your infrastructure.


aurelleb:
this looks mostly good to me, thank you for submitting this!

> If you’d prefer the spec and COPR workflow to live under a different path (e.g. packaging/fedora/), I can move them.

I think that would be better indeed, in order to avoid bloating the root of the repository. I have no problem with versioning the cmark patch but it should not be at the root of the repo IMO

quadratech188:
Hi, why was this closed?

aurelleb:
image
 @dacrab ?

PR #745, opened 2025-11-18
test: file-indexer

AntoineGS:
These were mostly just for me to get some things under test throughout the changes I made recently, and I figure they might be helpful if we move this to a service depending on the amount of code change that is needed.
I can adapt them to the service-model once you have something for it, as there are some Qt dependencies in some of these tests.
Feel free to cherry pick the tests or not use them :)


AntoineGS:
Hmm not sure why that test is failing I'll review it tomorrow!

AntoineGS:
I am closing the PR for now, I'll wait until the file indexing is its own thing before going back to it as it will likely change things anyways (I never was able to figure out the failing test on the ci, it passes on my machine 🤷)

PR #747, opened 2025-11-18
docs: add typedoc plugin to expand enum members for type aliases

zSoulweaver:
TypeDoc plugin as discussed on Discord briefly last week.

Adds a TypeDoc plugin to resolve type aliases that contain aliases enums to provide better documentation without needing to export said enums.

AI Disclaimer: Was built primarily using both Qwen Code and GPT 5

### Feedback Wanted

Plugin is built in Javascript since TypeDoc can only load JS files so we do lose some type safety from TS. We could use `@ts-check` and use JSDoc to type out the file, not exactly ideal, but it would work to provide some types. Alternative is to build the plugin with Typescript and require it to be built prior to doc generation.

Another thing I wasn't sure on was placement of the plugin file, I've just got it sitting in the root directory at the moment, but I'm sure it could likely live somewhere specific.

### What does it do?
- Plugin will retrieve all TypeAlias kinds in the project
- Ensure that TypeAlias is an referencing an enum before continuing
  - Since the enums we're interested in are usually not exported, if we can't find them in the TypeDoc solution, we need to fallback to Typescript to parse the file for us
  - If the Typescript pathway is chosen, we will generate the enum information from the data Typescript can provide us in the `buildEnumInfoFromAST` function
- Finally, we create the enum documentation by manually pushing the markdown into the Typedoc comment summary

### Gotchas
This will not generate a 1:1 page of an enum like you would see on generated pages for exported enums. This is done so that Typedoc will still show the kind being a "Type Alias" and not an enum, which it is not. The idea is to simply expand the enum members into the type alias documentation. Due to this, our generated enum page will look different to a true enum page.

Generating the "Defined in [file]" link is more difficult than expected and needs a bit of work to get going. The plugin is already complex enough (in my opinion) just to generate some simple enum members, so I decided against this after stuffing around with it for awhile. If we really wanted this, I'm happy to keep exploring this.

I haven't extensively tested this plugin, but I couldn't see any weird outputs in the generated docs. In case of any weird outputs or aliases that don't need be expanded there is a new `@noExpand` tag that can be added in the JSDoc to skip over the expansion.

### Final Remarks
This plugin works fine (if not better) with the `typedoc-plugin-markdown` to generate markdown files from the documentation. This unlocks the ability to use third party documentation frameworks like Vitepress which is also provided by the same folk who developed `typedoc-plugin-markdown`.

Finally, is this the right approach to this issue? No idea.


aurelleb:
This is a fine solution, much better than what we had (nothing) so that's still really helpful. Let's go with that.

I didn't know about vitepress and it looks really really cool, we might eventually switch to that.

Thank you!

PR #754, opened 2025-11-19
nix: remove flake-utils, use overlay's pkgs to call vicinae

0xda157:
Overlays should be called with `prev` so that the consumer's other overlays will apply to vicinae's dependencies. Related to https://github.com/vicinaehq/vicinae/pull/507, this will causes user's of the overlay with a different nixpkgs to have a different derivation, which will cause the vicinae in the cachix not to be used. This is desirable because as previously stated overlays should respond to the consumer's nixpkgs.

flake-utils is an unnecessary dependency which can be removed while keeping the functionality of having a system input the consumer can override. The system input can also be removed at the expense of users on platforms other than those explicitly supported losing the ability to use this flake (I've opted not to remove it).

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,23 +1,5 @@
 {
   "nodes": {
-    "flake-utils": {
-      "inputs": {
-        "systems": "systems"
-      },
-      "locked": {
-        "lastModified": 1731533236,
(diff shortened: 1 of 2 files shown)


aurelleb:
thanks!

PR #756, opened 2025-11-20
nix/mkVicinaeExtnesion: simplify, follow conventions

0xda157:
Moving it to packages and using callPackage because it is more convenient to not need to pass pkgs in. Deprecating name in favor of pname and version because the end user might want to set the version and it's consider bad practice to not set the version and to set name directly in Nixpkgs.

effects for users
- derivation name(s) will change from `name` to `name-0`
- they will get 3 warnings:
  ```
  evaluation warning: vicinae: accessing mkVicinaeExtension from flake top level is deprecated, use packages.<system>.mkVicinaeExtension
  evaluation warning: mkVicinaeExtension: calling with pkgs is deprecated and no longer necessary
  evaluation warning: mkVicinaeExtension: name has been replaced with pname
  ```

difficulty of migrating existing configs:
```diff
- (inputs.vicinae.mkVicinaeExtension.${pkgs.stdenv.hostPlatform.system} {
-   inherit pkgs;
-   name = "nix";
+ (inputs.vicinae.packages.${pkgs.stdenv.hostPlatform.system}.mkVicinaeExtension {
+   pname = "nix";
```

All of this has been tested locally before and after migrating my config to the new format.

### DIFF
--- a/nix/mkVicinaeExtension.nix
+++ b/nix/mkVicinaeExtension.nix
@@ -4,15 +4,18 @@
   importNpmLock,
 }:
 {
-  name,
+  pname ? lib.warn "mkVicinaeExtension: name has been replaced with pname" name,
+  version ? "0",
   src,
-  # Keep to prevent breaking api and provide warning.
+  # Keep to prevent breaking api and provide warnings.
   # TODO: throw after 26.05, remove after 26.11 NixOS release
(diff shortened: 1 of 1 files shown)


aurelleb:
thank you for your PR, can you fix the conflict please?

0xda157:
> thank you for your PR, can you fix the conflict please?

done, although https://github.com/vicinaehq/vicinae/pull/766 should be merged first.

PR #765, opened 2025-11-21
fix: Allow Enter (Numpad's Return) to execute Return actions

### LINKED ISSUE #734: Numpad enter key not working
**System information**

- Version: v0.16.6 (dc2485f3a)
- Provenance: binary_release
- OS: CachyOS
- QT Platform: wayland
- DE: GNOME/Wayland

**Describe the bug**

The numpad Enter key does not trigger the same action as the main (alphanumeric) Enter key inside Vicinae.  

Main Enter correctly confirms and runs the selected item in the UI, but numpad Enter does nothing, even when both keys are mapped by the system to produce the same keycode.

**To Reproduce**

1. Start Vicinae.
2. Invoke the main search input.
3. Type any query so that a list of results appears.
4. Use the arrow keys to highlight a result.
5. Press the **main Enter** key  
   - The selected item is executed as expected.
6. Now repeat steps 2–4, but this time press **numpad Enter**  
   - Nothing happens, and the selected item is not executed.

This is reproducible with default settings, no custom keybindings, and with the numpad enabled (Num Lock on).

**Expected behavior**

Pressing numpad Enter should behave exactly the same as the main Enter key in Vicinae.  

In particular, when a result is highlighted, both main Enter and numpad Enter should confirm and run the selected item.

**Screenshots**

Not applicable, since this is a key handling issue rather than a visual glitch.

**Additional context**

- Environment: GNOME on Wayland (CachyOS, Qt platform set to `wayland`).
- I tested mapping numpad Enter to behave like normal Enter at the system level using `keyd`:
  - `KPENTER = enter`
  - `showkey -a` reports the same character code for both main Enter and numpad Enter after this mapping.
- Despite that, Vicinae still treats them differently. Main Enter works, numpad Enter does not.
- This suggests Vicinae might be handling only one of the Enter variants exposed by Qt (for example `Qt::Key_Return`) and not the keypad variant (for example `Qt::Key_Enter`), or otherwise distinguishing between the two in key handling.

Happy to run any debug build or provide extra logs if needed.

prettysunflower:
Nyallo!

This pull request intends to close #734 

> Quick reminder: Return -> Enter/Return on the "main" keyboard, Enter -> Enter/Return on the numpad

**Why**
Users may expect the Enter key to produce the same actions as the Return key, as seen in the mentioned issue, and my personal experience using the program (my muscle memory is using the Enter key, and nothing happens)

**How**
The keyboard shortcuts for actions are being compared by [`Shortcut::operator==`](https://github.com/vicinaehq/vicinae/blob/d7e98b4447d5c7df88e3e67c5ddfbae56730d341/vicinae/src/lib/keyboard/keyboard.cpp#L287).

Users may want to be able to use the Enter key like the return key, and because:
- The generic `enter` Shortcut is configured with `Qt::Key_Return` [(link)](https://github.com/vicinaehq/vicinae/blob/d7e98b4447d5c7df88e3e67c5ddfbae56730d341/vicinae/src/lib/keyboard/keyboard.hpp#L15) and
- `isBoundTo` remap the Qt::Key_Enter to Qt::KeyReturn with the original modifiers [(link)](https://github.com/vicinaehq/vicinae/blob/d7e98b4447d5c7df88e3e67c5ddfbae56730d341/vicinae/include/ui/action-pannel/action.hpp#L49), and
- `Qt::Key_Enter` comes with `Qt::KeypadModifier`, and `Qt::Key_Return` doesn't, the numpad's Enter couldn't be used.

This commit/pull request fixes this by flipping the `Qt::KeypadModifier` flag in the shortcut modifiers, allowing the action to be run with the enter key.

### DIFF
--- a/vicinae/include/ui/action-pannel/action.hpp
+++ b/vicinae/include/ui/action-pannel/action.hpp
@@ -5,6 +5,7 @@
 #include <qcontainerfwd.h>
 #include <qevent.h>
 #include <qlogging.h>
+#include <qnamespace.h>
 #include <qtmetamacros.h>
 
 class AppWindow;
@@ -46,16 +47,7 @@ public:
   bool isBoundTo(const QKeyEvent *event) {
     if (event->key() == Qt::Key_Enter) {
       qDebug() << "remapping numpad enter to return";
-
-      /*
-       * The action is being compared with the modifier in Shortcut::operator==.
-       * Users may want to be able to use the numpad's Enter key like the (non-numpad) return key. (Issue 734)
-       * For that to work, we need to remove the KeypadModifier flag from the modifiers list.
-       */
-      Qt::KeyboardModifiers modifiers = event->modifiers();
-      modifiers ^= Qt::KeypadModifier;
-
-      return isBoundTo(Keyboard::Shortcut(Qt::Key_Return, modifiers));
+      return isBoundTo(Keyboard::Shortcut(Qt::Key_Return, event->modifiers() ^= Qt::KeypadModifier));
     }
 
     return isBoundTo(Keyboard::Shortcut(event));


aurelleb:
That did the trick, thanks! I rewrote it in a slightly more succinct way tho.

prettysunflower:
No problem! Thank you very much! :)

PR #703, opened 2025-11-12
Add ActionPanel Submenu (and default title for Open In Browser action)

rithvikvibhu:
## Action Panel Sub Menu

closes #706.

extension: `wikipedia` (can also see title says Open in Browser)
image

extension: `timezone-converter`
image

boilerplate:
image

PR #768, opened 2025-11-22
Actionpanel submenu with conditional updates

### LINKED ISSUE #706: ActionPanel.Submenu support
```
  // Note: submenu are currently not implemented and a good implementation will require
  // good thinking on how to handle state (especially regarding extensions).
  virtual bool isSubmenu() const { return false; }
```

Hi team,

I noticed that Submenu is available in the API, but it doesn’t seem to be implemented yet. I couldn’t find any existing issue about it, so I’m creating this one to ask about the current plan.

If I missed something or there’s already progress elsewhere, feel free to close this. Thanks!

rithvikvibhu:
Same as #703 but with a few more commits ([view only diff](https://github.com/rithvikvibhu/vicinae/compare/actionpanel-submenu...rithvikvibhu:vicinae:actionpanel-submenu-stateful))

closes #706.

extension: `github`

image


aurelleb (approved):
I tested it and it seems to work pretty well!
I pushed a fix so that the action panel view stack is properly reset every time it is closed, which wasn't the case before.

Now about the implementation, there are many, many things I don't like about this, as I expected. I'm not going to pinpoint every single thing because most of it is just because you had to workaround an action panel that clearly wasn't built with multiple views in mind.

I think I'm okay with merging this until I can rethink the way this entire thing works.

aurelleb (approved):
*was sent 2 times for some reason*

aurelleb:
no idea why this test started failing in the CI, it's all right locally.

rithvikvibhu:
Passed this time `¯\_(ツ)_/¯`

PR #769, opened 2025-11-22
Markdown: render inline content next to images

rithvikvibhu:
When there is an image and other content in a markdown html_block, then only the image is rendered and the rest of the content ignored.

That meant some inline text like this would not render:
```
const test =
    '\n<img src="https://howlongtobeat.com/games/65945_Hollow_Knight_Silksong.jpg?width=760" width="150" />\n# Hollow Knight: Silksong \nAs the lethal hunter Hornet, adventure through a kingdom ruled by silk and song! Captured and taken to this unfamiliar world, prepare to battle mighty foes and solve ancient mysteries as you ascend on a deadly pilgrimage to the kingdom’s peak.\n\n## Platform\nUpdated: 16 Mins Ago\n    ';
```

Its AST looks like:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE document SYSTEM "CommonMark.dtd">

<document xmlns="http://commonmark.org/xml/1.0">
  <html_block>&lt;img src=&quot;https://howlongtobeat.com/games/65945_Hollow_Knight_Silksong.jpg?width=760&quot; width=&quot;150&quot; /&gt;
# Hollow Knight: Silksong
As the lethal hunter Hornet, adventure through a kingdom ruled by silk and song! Captured and taken to this unfamiliar world, prepare to battle mighty foes and solve ancient mysteries as you ascend on a deadly pilgrimage to the kingdom’s peak.</html_block>
  <heading level="2">
    <text>Platform</text>
  </heading>
  <paragraph>
    <text>Updated: 16 Mins Ago</text>
  </paragraph>
</document>
```

With this new change, all non-img content is collected and parsed and rendered as markdown again.

I did consider parsing markdown around each img separately, but this is an edge case and even raycast does not render markdown, so figured it's okay.

example: `how-long-to-beat`
image

PR #780, opened 2025-11-24
markdown: render html block with libxml

rithvikvibhu:
This replaces the regex based `` parser with libxml. And also renders markdown for text nodes inside the html block. Supersedes #769.

example: `how-long-to-beat`
image

extension: `store.vicinae.systemd`
image

boilerplate: blue square (img with src that includes a `<`, invalid xml, valid html) still renders fine
image


aurelleb:
@ArjixWasTaken @cilginc @jaredallard this adds a new direct dependency on `libxml2`. It's worth noting that QT already depends on it so the vicinae binary links against it already, at least on my system. Still worth adding explicitly in order to make sure the headers are available

PR #808, opened 2025-12-03
feat: add COPR packaging and automation for vicinae

dacrab:
## Summary

This PR adds COPR packaging support and build automation for vicinae so that Fedora users can enable a COPR repo and install vicinae via the package manager.

> **Note**: This is a resubmission of #733 which was closed when I accidentally deleted my fork.

### 1. RPM spec file (`packaging/fedora/vicinae.spec`)
A Fedora-style spec file that builds vicinae as an RPM using upstream sources.

### 2. GitHub Actions workflow (`.github/workflows/copr-release.yml`)
Triggers COPR builds on tagged releases (`v*`) or manual dispatch.

### 3. Helper script (`scripts/copr-build.sh`) + Makefile target
Run `make copr-build` to build SRPM and submit to COPR locally.

### 4. cmark-gfm patch (`packaging/fedora/cmark-gfm-no-git.patch`)
Allows offline builds in RPM chroots by using tarball instead of git fetch.

### 5. Build fix: `#ifdef WAYLAND_LAYER_SHELL` guards
Guards LayerShellQt includes so the build works with `-DWAYLAND_LAYER_SHELL=OFF`.

### 6. `.gitignore` updates
Ignores `/.build`, `/.rpm`, `/.srpm` directories.

## Testing

- Successfully built with `WAYLAND_LAYER_SHELL=OFF`
- Successfully built with `WAYLAND_LAYER_SHELL=ON`
- COPR build submitted: https://copr.fedorainfracloud.org/coprs/build/9863352

## Impact

No changes to core vicinae behavior - only adds packaging/CI infrastructure.


ArjixWasTaken:
~~is COPR basically the same thing as the AUR?~~

gemini says that the difference is that fedora's cloud builds the package, not the end user

quadratech188:
```
quadratech@fedora:~$ sudo dnf install vicinae
Updating and loading repositories:
Repositories loaded.
Failed to resolve the transaction:
Problem: conflicting requests
  - nothing provides libxdgpp.so()(64bit) needed by vicinae-0.16.6-1.fc43.x86_64 from copr:copr.fedorainfracloud.org:killcrb:vicinae
You can try to add to command line:
  --skip-broken to skip uninstallable packages
```

I ran into this issue when I was playing around with packaging, and it happens with yours too.

Basically: the default `%cmake` macro specifies `-DBUILD_SHARED_LIBS:BOOL=ON`, which makes xdgpp build as a shared library and makes vicinae depend on it, but xdgpp installation was disabled. (#509)

quadratech188:
Why did you decide to turn `WAYLAND_LAYER_SHELL` off? Does it cause problems for other compositors?

Also you can [make COPRs depend on other COPRs](https://discussion.fedoraproject.org/t/copr-that-depends-on-another-copr/45653/2), so maybe `cmark-gfm-no-git.patch` isn't necessary.

dacrab:
> Why did you decide to turn `WAYLAND_LAYER_SHELL` off? Does it cause problems for other compositors?
> 
> Also you can [make COPRs depend on other COPRs](https://discussion.fedoraproject.org/t/copr-that-depends-on-another-copr/45653/2), so maybe `cmark-gfm-no-git.patch` isn't necessary.


Thanks for the feedback!

WAYLAND_LAYER_SHELL: You're right - I've now enabled it. It was causing build failures in the COPR mock chroot earlier, but I just retested and it builds fine now: https://copr.fedorainfracloud.org/coprs/build/9908832

COPR dependencies: I went with the tarball + patch approach to keep the build self-contained and avoid depending on external COPRs that might become unmaintained. But if there's a well-maintained COPR for cmark-gfm you'd recommend, happy to consider switching.

quadratech188:
here's one that I quote-unquote "maintain": (cmark-gfm hasn't seen a commit in 2 years anyway)
https://copr.fedorainfracloud.org/coprs/quadratech188/cmark-gfm

aurelleb:
is fetching dependencies during build possible for the copr build? We recently added a dependency on the glaze library (https://github.com/stephenberry/glaze) which is needed at build time only. Currently it's getting automatically fetched by default unless manually disabled.

quadratech188:
Manually getting stuff over the network is *possible* but not recommended. Since `glaze` doesn't exist in normal Fedora repos
> COPR dependencies: I went with the tarball + patch approach to keep the build self-contained and avoid depending on external COPRs that might become unmaintained. But if there's a well-maintained COPR for cmark-gfm you'd recommend, happy to consider switching.

So the proper way is to either bundle glaze with source, or create a COPR for it.

quadratech188:
But come to think of it, we download hundreds of npm packages from the internet anyway!

aurelleb:
@quadratech188 if that's possible, I think it's better to just turn on the flag that fetches it from github (that's the default for glaze, and cmark-gfm can do it too but there is the copr). It's much simpler to deal with for me, unless someone wants to make a new copr for glaze or something. I'm completely clueless when it comes to fedora packaging so I'm not going to pretend I understand what this entails exactly

quadratech188:
@dacrab Would you mind triggering a build (after updating the version) to see if everything still works?

quadratech188:
glaze seems to install headers and CMake files in /usr/share/ and /usr/include, @aurelleb is there a way to disable that? (I know next-to-nothing about CMake)

aurelleb:
@quadratech188 what do you mean? You want to change the install location? It's a header only library so all it does is install headers.

quadratech188:
The headers aren't required for vicinae to work, they don't need to exist. I *can* tell the packaging program to not include them in the package, but ideally they shouldn't be installed at all.

aurelleb:
oh I didn't realize it was installing them when you install vicinae, that's a mistake. I 'm gonna fix this.

aurelleb:
fixed on main

dacrab:
@quadratech188 

Fixed! The issue was that `%cmake` macro sets `BUILD_SHARED_LIBS:BOOL=ON` by default, which made xdgpp build as a shared library. Since xdgpp installation is disabled, this created a missing `libxdgpp.so()` runtime dependency.

I've fixed it by explicitly setting `-DBUILD_SHARED_LIBS:BOOL=OFF` in the `%cmake` call, which forces xdgpp to build statically and be linked into vicinae instead of creating a separate shared library.

The build now succeeds: https://copr.fedorainfracloud.org/coprs/build/9941682



Done! I've updated the version to 0.17.1 and triggered a build. The build succeeded: https://copr.fedorainfracloud.org/coprs/build/9941682

I also fixed a few issues that came up:
- Fixed the xdgpp dependency issue (see my reply above)
- Fixed a CMake generator conflict (removed `-G Ninja` since `%cmake` uses Unix Makefiles by default)
- Removed glaze headers/CMake files from the package (they're build-time only, so I added cleanup steps in `%install` to remove them)

Everything is working now! 🎉

dacrab:
@aurelleb 

Glaze is working fine! It gets fetched automatically from GitHub during the build via CMake FetchContent (default behavior), which is acceptable since we're already downloading npm packages during the build anyway.

I did have to add a cleanup step in the `%install` section to remove the glaze headers and CMake files that were being installed (since they're build-time only and shouldn't be packaged). This is a temporary workaround until the fix on main makes it into a tagged release.

The build succeeded: https://copr.fedorainfracloud.org/coprs/build/9941682

aurelleb:
the headers are no longer installed so the cleanup is no longer required

dacrab:
> the headers are no longer installed so the cleanup is no longer required

Removed the cleanup step - thanks for the update! The spec file is now cleaner without the unnecessary cleanup code.

PR #810, opened 2025-12-04
build: do not link against minizip statically for portable builds

aurelleb:
@ArjixWasTaken please add `minizip-ng` to the `vicinae-bin` package dependencies. No change is needed for the other packages.


ArjixWasTaken:
Done

PR #818, opened 2025-12-05
fix(markdown-renderer): fit image in render window

Gareck-OG:
When rendering images that dont have the correct height and width images could get rendered squashed. Now render looks how it is supposed to look like. Fix issue #819 
## Comparison for rendering
```typescript
<List.Item.Detail markdown={`![](${ImagePath})`} />
```

### Before:
image

### After:
image

## Previous workaround
```typescript
<List.Item.Detail
markdown={`<img height="320" src="${previewPath}" />`}
/>
```
**NOTE:** Any other workaround via HTML styling I tried didnt work.

### DIFF
--- a/extra/extension-boilerplate/src/simple-detail.tsx
+++ b/extra/extension-boilerplate/src/simple-detail.tsx
@@ -59,6 +59,14 @@ yellow rectangle with text (data:image/svg+xml base64):
 markdown format image (data:image/png;base64 with raycast-* size):
 
 ![Logo](${DATA_URI_PNG_BASE64}?raycast-width=50&raycast-height=50)
+
+markdown format image with style attribute (no unit):
+
+<img src=${DATA_URI_PNG_BASE64} style="width:50; height:50">
+
+markdown format image with style attribute (px):
+
+<img src=${DATA_URI_PNG_BASE64} style="width:50px; height:50px">
 `;
 
 const table = `


vicinae/src/ui/markdown/markdown-renderer.cpp
```diff
@@ -592,6 +600,37 @@ void MarkdownRenderer::insertHtmlImage(xmlNode *node) {
         if (!value.isEmpty()) { imgSize.setWidth(value.toInt()); }
       } else if (name == "height") {
         if (!value.isEmpty()) { imgSize.setHeight(value.toInt()); }
+      } else if (name == "style") {
```
aurelleb: I don't think I want to support inline styling like this, are some Raycast extensions expecting this? If that's not the case please remove this part.
The reason I don't want it is because supporting this suggests to the user we support inline styling in general which is realistically never going to be the case.

Gareck-OG: @aurelleb 
Ill remove it. I just added it since i normally default to inline styling when adjusting height and width. I checked and only a handfull (have spotted 5) of Raycast extentions use inline styling for images.

aurelleb: if some do then I guess it makes sense to have this limited subset supported

aurelleb:
since some extensions make use of this it is desirable to support it. I also added some more examples using the attribute to the boilerplate

PR #828, opened 2025-12-08
fix: use layer top instead of overlay

Aleksanaa:
To avoid obstructing input method candidate box.

Without this patch, fcitx5 always appears behind the popup (only when it's in wayland mode, it works if it is loaded as QT_IM_MODULE):

2025-12-08T21:13:18,865865434+08:00

With this patch:

2025-12-08T21:19:31,934612063+08:00

### DIFF
--- a/vicinae/src/services/file-chooser/native/native-file-chooser.cpp
+++ b/vicinae/src/services/file-chooser/native/native-file-chooser.cpp
@@ -26,7 +26,7 @@ NativeFileChooser::NativeFileChooser(QObject *parent) : AbstractFileChooser(pare
 
     m_dialog.createWinId();
     if (auto lshell = Shell::Window::get(m_dialog.windowHandle())) {
-      lshell->setLayer(Shell::Window::LayerOverlay);
+      lshell->setLayer(Shell::Window::LayerTop);
       lshell->setScope(Omnicast::APP_ID);
       lshell->setScreenConfiguration(Shell::Window::ScreenFromCompositor);
       lshell->setKeyboardInteractivity(Shell::Window::KeyboardInteractivityExclusive);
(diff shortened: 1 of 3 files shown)


aurelleb:
thanks!

saatvik333:
But doesn't this defeats the usability when working on full-screen apps and user wants to open something?

Aleksanaa:
> l-screen apps and user wants to open something?

Do you already have a full-screen app that displays on top of vicinae (layer top)? I didn't see any in my testing

saatvik333:
> > l-screen apps and user wants to open something?
> 
> Do you already have a full-screen app that displays on top of vicinae (layer top)? I didn't see any in my testing

https://github.com/user-attachments/assets/76a275b1-df6a-4e45-8f03-a3d087eb1eca

OS: Cachy
WM: Niri
Vicinae: 0.16.14 vicinae-bin

Aleksanaa:
In Niri, ZWLR_LAYER_SHELL_V1_LAYER_TOP is hidden behind windows:

> When a [full-screen](https://github.com/YaLTeR/niri/blob/main/docs/wiki/Fullscreen-and-Maximize.md) window is active and covers the entire screen, it will render above the top layer, and it will be prioritized for keyboard focus. If your launcher uses the top layer, and you try to run it while looking at a full-screen window, it won't show up. Only the overlay layer will show up on top of full-screen windows.

https://github.com/YaLTeR/niri/blob/main/docs/wiki/Layer%E2%80%90Shell-Components.md

While in Hyprland, ZWLR_LAYER_SHELL_V1_LAYER_TOP is rendered on top of windows:

https://github.com/hyprwm/Hyprland/blob/main/src/render/Renderer.cpp#L1032

I don't think there is a particularly good solution here. Perhaps it would be better to let users choose which layer they use.

aurelleb:
this will be made configurable soon

lluisaliano:
Has this parameter been made configurable? Can't find a way to make it work on niri

aurelleb:
yes check the example config @lluisaliano

PR #829, opened 2025-12-08
feat: arrange browse-app-view by score

Aleksanaa:
Similar to how we arrange root item list, and also similar to rofi drun behavior. I find it quite convenient to browse a list sorted by recentness/frequency when I don't want to type a few characters to search.

### DIFF
--- a/vicinae/src/extensions/system/browse-apps/browse-apps-view.cpp
+++ b/vicinae/src/extensions/system/browse-apps/browse-apps-view.cpp
@@ -7,6 +7,7 @@
 #include "ui/omni-list/omni-list.hpp"
 #include "service-registry.hpp"
 #include "services/app-service/app-service.hpp"
+#include "services/root-item-manager/root-item-manager.hpp"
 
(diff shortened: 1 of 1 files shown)


aurelleb:
I'm generally not a fan of how this is done, the browse apps view should not directly get and set the metadata like this. 

Also, the app service should not have to know about the root item manager.

However, I agree that it makes sense to have the apps in this menu be sorted by frequency as well.

I think this might be solved by me implementing provider-scoped deeplinks like:
vicinae://apps
vicinae://shortcuts
vicinae://extensions/<id>

Aleksanaa:
> I'm generally not a fan of how this is done, the browse apps view should not directly get and set the metadata like this.
> 
> Also, the app service should not have to know about the root item manager.

Then I think the extensions should also have a formal interface to write to global visit statistics database. Otherwise, even if an app is selected a thousand times in browse-app-view, it will not be ranked higher in suggestions of root menu, which does not make sense.

PR #833, opened 2025-12-09
feat: allow alphabetical sorting for all items & allow calc operations inside of calculator history module & detect uwsm session

mkbula:


aurelleb:
Also forgot to note this, but you should rebase on main to take the new config changes in

PR #854, opened 2025-12-17
fix: fix cachix being run on prs

schromp:
push to cachix shouldnt run when opening a pr.
this breaks the cache, obviously


aurelleb:
this has been broken from the start then, damn. Thanks!

PR #855, opened 2025-12-17
Fix: Ensure clipboard text compatibility with X11 terminals

### LINKED ISSUE #516: Text copied from Vicinae cannot be pasted into Neovim or terminal under X11
### System information

* **OS:** Void Linux (X11)
* **Terminal:** st (suckless terminal)
* **Clipboard manager:** Vicinae

### Description

The previous issue with text copied **from st** not appearing in the system clipboard has been **resolved** — thank you! 🎉

However, there’s a new issue:
Text copied **from Vicinae** cannot be pasted into **Neovim** or even back into **st** itself.
Copying from Vicinae works fine in other applications (such as browsers or graphical editors), but pasting into terminal-based programs doesn’t work — nothing is inserted.

### To Reproduce

1. Copy some text from Vicinae.
2. Open Neovim or st under X11.
3. Try to paste the text (using `"+p`, `Ctrl+Shift+V`, or middle-click).
4. Nothing is pasted — the clipboard content isn’t recognized.

### Expected behavior

Text copied from Vicinae should be available through the **CLIPBOARD** selection and pasteable into both GUI and terminal-based applications (including Neovim and st).

### Additional context

The previous `st` clipboard issue is fixed — thank you again!
Now the problem seems reversed: copying *from* Vicinae doesn’t propagate correctly to applications relying on X11 clipboard selections inside the terminal.

Tinkerbells:
### Description
This PR addresses an issue where copied text from Vicinae could not be pasted into certain X11 terminal emulators (like `st`, `xterm`, or `alacritty` in specific configurations).

### The Problem
In minimalist X11 environments (e.g., Void Linux with `dwm`), there is often no global clipboard manager to normalize MIME types. When Vicinae sets clipboard data using generic `setData("text/plain", ...)` calls, the Qt X11 platform plugin might not always expose the standard `UTF8_STRING` atom. As a result, terminals that strictly adhere to X11 selection standards see the clipboard as empty or incompatible.

### The Solution
The fix involves using `QMimeData::setText()` specifically for text-based MIME types. When `setText()` is called, the Qt backend automatically handles the negotiation and population of standard X11 atoms (`UTF8_STRING`, `STRING`, `TEXT`), ensuring broad compatibility across all X11 clients.

### Verification
- **Environment:** Void Linux, X11, dwm.
- **Before:** Text copied from Vicinae was invisible to `st` and required a proxy like `copyq` to work.
- **After:** Text is successfully pasted into `st` and other terminals natively.
- **Regression Check:** No impact on Wayland sessions, as `setText` remains a standard and safe Qt method for text handling.

Closes #516

### DIFF
--- a/vicinae/src/services/clipboard/clipboard-service.cpp
+++ b/vicinae/src/services/clipboard/clipboard-service.cpp
@@ -543,7 +543,11 @@ bool ClipboardService::copySelection(const ClipboardSelection &selection,
         mimeData->setImageData(img);
       }
     } else {
-      mimeData->setData(offer.mimeType, offer.data);
+      if (Utils::isTextMimeType(offer.mimeType)) {
+        mimeData->setText(QString::fromUtf8(offer.data));
+      } else {
+        mimeData->setData(offer.mimeType, offer.data);
+      }
     }
   }
 


aurelleb:
lgtm, thanks!

PR #858, opened 2025-12-18
nix: simplify derivation based on nixpkgs

whispersofthedawn:
Simplify the derivation based on the version that is upstream in Nixpkgs. See https://github.com/NixOS/nixpkgs/blob/a008f4366465ac03ac6f79af0edee60959aaa685/pkgs/by-name/vi/vicinae/package.nix for the version this is based on. This allows avoiding needless dependencies and specifying build steps where not needed, since the built-in Nixpkgs machinery for CMake and Ninja can handle it.

Also note that this does not include `glaze` (which'll be in the next version) since that change hasn't been committed yet, but I can add that once that commit is made.

This supersedes #245, which can be closed if/after this is merged.


aurelleb:
thanks!

PR #859, opened 2025-12-19
config: move provider alias/preferences to config file

aurelleb:
BREAKING: breaks most configurations


aurelleb:
@cilginc @ArjixWasTaken @jaredallard this adds a new dependency on https://github.com/stephenberry/glaze v6 (only required at build time)

PR #860, opened 2025-12-20
fix: set raw image data for clipboard images, resolves #826

### LINKED ISSUE #826: Copying an image to clipboard from clipboard history doesn't work.
**System information**

- Version: v0.16.11 (05cbd00db)
- Provenance: binary_release
- OS: Arch Linux - 
- QT Platform: wayland
- DE: GNOME/Wayland

**Describe the bug**

When trying to copy an image that exists in the clipboard history inside Vicinae, it doesn't do that; it gives the error `Failed to copy to clipboard.`, and the logs give: `GnomeClipboardServer: Empty data for format "application/x-qt-image"`.

**To Reproduce**

- Copy an image.
- Copy some text.
- Access the clipboard from Vicinae, then go to the image and press `Shift + Enter` to copy, and an error will appear: `Failed to copy to clipboard`.

**Expected behavior**

The image is copied to the clipboard, and Vicinae closes, just like what happens with text entries.

**Screenshots**

Image

**Additional context**

I think it's related to GNOME's clipboard server compatibility, because it works on other DEs.

dagimg-dot:
- this is because Gnome D-Bus needs the raw data

### DIFF
--- a/vicinae/src/services/clipboard/clipboard-service.cpp
+++ b/vicinae/src/services/clipboard/clipboard-service.cpp
@@ -540,9 +540,10 @@ bool ClipboardService::copySelection(const ClipboardSelection &selection,
         qWarning() << offer.mimeType << "could not be converted to valid image format";
         mimeData->setData(offer.mimeType, offer.data);
       } else {
-        mimeData->setData(offer.mimeType, offer.data);  // Raw bytes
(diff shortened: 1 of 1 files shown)


aurelleb:
thanks for the fix! As a general rule please avoid comments unless they are really adding something useful

Issue #852, opened 2025-12-16
Vicinae toggle shows unexpected application icon on panel


jsicas:
**System information**

- Version: v0.16.14 (04608039d)
- Build info: GCC 15.2.0 - Release - LTO
- Provenance: appimage
- OS: Fedora Linux 43 (Workstation Edition) - 43
- QT Platform: wayland
- DE: GNOME/Wayland

**Describe the bug**

When the Vicinae toggle window is opened, Vicinae appears as a running application icon on the panel.

**To Reproduce**

1. Install Vicinae using the official installation script
1. Enable the Vicinae systemd **user** service
1. Configure a keyboard shortcut to trigger the Vicinae toggle
1. Press the shortcut to open Vicinae

**Expected behavior**

Vicinae should not appear as an application icon on the panel when opening the toggle window.

**Screenshots**

<img width="1600" height="900" alt="Image" src="https://github.com/user-attachments/assets/2d91a087-968e-4f7a-9940-4088f530a041" />

**Additional context**

I'm using Dash to Panel extension.

he3als:
This [might be fixed](https://github.com/vicinaehq/vicinae/pull/868) in the latest release?

jsicas:
I reinstall it, but It's still not working.

andypiper:
Same issue here.

PR #868, opened 2025-12-21
fix(.desktop): disable StartupNotify

he3als:
Disables StartupNotify to prevent a lingering icon in the taskbar of KDE Plasma when toggling Vicinae.

# Before

https://github.com/user-attachments/assets/d583b45a-b3c0-4c5c-a510-e72089d1b798

# After

https://github.com/user-attachments/assets/3d8d72e2-20ec-4304-b0f9-1f55d00c496e

### DIFF
--- a/extra/vicinae.desktop
+++ b/extra/vicinae.desktop
@@ -10,6 +10,7 @@ Categories=Utility;Accessibility;
 NoDisplay=false
 Actions=open;close;toggle
 StartupWMClass=vicinae
+StartupNotify=false
 
 [Desktop Action open]
 Name=Open Vicinae Window


aurelleb:
oh that's what it was then! I guess that will also fix it on gnome

he3als:
I presume so, apps like Spectacle (the Plasma screenshot app) also disable StartupNotify.

aurelleb:
that's great, thanks again :)

yrifl:
Screenshot From 2025-12-23 18-44-34

Not fixed on gnome using Dash to Dock extension, well it was a blank icon at first like on this issue https://github.com/vicinaehq/vicinae/issues/852, I managed to fixed it by modifying `vicinae.desktop` value of Icon to `/usr/local/lib/vicinae/vicinae.png`

Screenshot From 2025-12-23 18-47-41

The problem now is that I can't figured out a way to hide Vicinae from showing to Dash to Dock, even with your fix `StartupNotify=false`. I don't know if it is a compatibility problem with Dash to Dock, but let me know if you know a workaround. 

Thanks!

yrifl:
Just noticed, this fix only tested on KDE Plasma.

yrifl:
https://github.com/user-attachments/assets/9cda7cdf-3b91-4a44-a879-591eebc1ef16

I found a weird temporary workaround, I don't know if it would help but can you check what's causing this behavior? This works after opening General, Keybinds, or About settings of Vicinae and the problem is back when I open one of those settings again.

PR #877, opened 2025-12-23
fix(nix): Fix new version compatibility

### LINKED ISSUE #875: Nix home manager module generates the configuration file with the wrong name
**System information**

- Version: v0.17.1 (106fa7024)
- Provenance: nix
- OS: NixOS 26.05 (Yarara) - 26.05
- QT Platform: wayland
- DE: Hyprland/Wayland

**Describe the bug**

The Nix home manager module generates the configuration file at `~/.config/vicinae/vicinae.json`, but as of `v0.17.0`, the config should be at `~/.config/vicinae/settings.json`. I suspect the format will have changed as well, but I haven't checked yet.

**To Reproduce**

* Use the home manager module to configure any setting
* The setting will not be applied, and a config will be generated at `~/.config/vicinae/vicinae.json`

mochouaaaaa:
Update the following:
-  `vicinae.json` to `settings.json`
-  Copy the home-manager option configuration without breaking state again 

add https://github.com/vicinaehq/docs/pull/40

Fixes #875

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -11,15 +11,13 @@ let
   inherit (pkgs.stdenv.hostPlatform) system;
   vicinaePkg = self.packages.${system}.default;
 
-  settingsFormat = pkgs.formats.json { };
-  themesFormat = pkgs.formats.toml { };
+  jsonFormat = pkgs.formats.json { };
+  tomlFormat = pkgs.formats.toml { };
 in
 {
 
   options.services.vicinae = {
-    enable = lib.mkEnableOption "vicinae launcher daemon" // {
-      default = false;
-    };
+    enable = lib.mkEnableOption "vicinae launcher daemon";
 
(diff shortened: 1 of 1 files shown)


aurelleb:
thanks!

nix/module.nix
```diff
@@ -28,16 +26,44 @@ in
...
 
-    autoStart = lib.mkOption {
-      type = lib.types.bool;
-      default = true;
-      description = "If the vicinae daemon should be started automatically";
-    };
+    systemd = {
+      enable = lib.mkEnableOption "vicinae systemd integration";
+
+      autoStart = lib.mkOption {
```
0xda157: this is an unnecessarily breaking change, [lib.mkRenamedOptionModule](https://noogle.dev/f/lib/mkRenamedOptionModule) should have been used here.

PR #878, opened 2025-12-23
feat: autodetect uwsm

mkbula:
Detects active session of uwsm. If active defaults to "uwsm-app --" launch prefix. Doesn't override user settings.


aurelleb:
thanks!

PR #879, opened 2025-12-23
fix: Use parent path for Open in folder

### LINKED ISSUE #847: Opening parent directory of a file also opens that file
**System information**

- Version: v0.16.14 (04608039d)
- Provenance: binary_release
- OS: CachyOS - 
- QT Platform: wayland
- DE: KDE/Wayland

**Describe the bug**

When I choose the action `Open in folder`, it also opens the file.

**To Reproduce**

1. Search for a file
2. Open the action menu
3. Select `open in folder`
4. Notice that both the file manager and the file are opened

**Expected behavior**

Only the file manager with the parent directory of that file is opened.

**Screenshots**

https://github.com/user-attachments/assets/6ab4ecd4-ba31-496f-8001-c0fd4b562c1a

**Additional context**

Happens both with when pressing enter on the option and when clicking with mouse on the `open in folder` action.

quadratech188:
Closes #847.

### DIFF
--- a/vicinae/src/misc/file-list-item.hpp
+++ b/vicinae/src/misc/file-list-item.hpp
@@ -19,13 +19,12 @@ protected:
   ImageURL getIcon() const { return ImageURL::fileIcon(m_path); }
 
   static OpenFileInAppAction *createOpenInFolderAction(const std::filesystem::path &path,
-      std::shared_ptr<AbstractApplication> browser) {
+                                                       const std::shared_ptr<AbstractApplication> &browser) {
 
     // These file managers don't work correctly when they're passed a file path
     // We work around this by passing the parent folder path instead
-    static std::set<QString> exceptions = {
-      "org.kde.dolphin.desktop", "ranger.desktop"
-    };
+    static const std::set<QString> exceptions = {"org.kde.dolphin.desktop", "ranger.desktop"};
+
     if (exceptions.contains(browser->id())) {
       return new OpenFileInAppAction(path, browser, "Open in folder", {path.parent_path().c_str()});
     }


quadratech188:
This works, but could be better.

Currently we just open the parent directory, but we don't indicate the target *file* in any way.

Some file managers like dolphin and nautilus can take a `--select` option which is exactly what we need, but passing that parameter would require ignoring the `.desktop` file's `Exec=` field, tearing apart the whole `AbstractAppDatabase::launch` abstraction.

quadratech188:
One thought I had was having (either telling the user to make one, generating one, or just making one up in memory) a new desktop file where the `Exec=` field already looks like `dolphin --select %u`, and special-casing for that, Does that sound good?

aurelleb:
@quadratech188 that's the canonical way to do it but it's also very impractical. But there is no real way for vicinae to handle it all by itself since flags like --select do not obey any standard

aurelleb:
but even asking the user to make a custom one is not great, because it would require to be set as the default file browser to work, which may not be desirable here. If we want a really deep integration I guess we will have to try to detect what file manager it is and explicitly pass the right flags. That means we need a new abstraction for this

aurelleb:
After testing this I think we should probably add custom specialization logic or do something else, because this breaks the way things currently work (provided that the file manager supports a path to a file, which nautilus and others seem to do).

quadratech188:
> provided that the file manager supports a path to a file, which nautilus and others seem to do

Ok, nautilus and thunar supports it (I thought thunar doesn't for some reason). I guess we should just special-case dolphin.

quadratech188:
BTW I found this: https://www.freedesktop.org/wiki/Specifications/file-manager-interface
sad thing thunar doesn't support it

EDIT: thunar does support it; I was looking at the wrong repository from 14 years ago. So using this could also be feasible.

EDIT2: I don't think this is feasible. dolphin and nautilus reuse windows whenever possible, and dolphin doesn't even change the window state to urgent - so if I had a dolphin window on another workspace I get zero feedback on whether the file was opened.

EDIT3: wait, firefox manages to set dolphin's window state to urgent. hang on

quadratech188:
> but even asking the user to make a custom one is not great, because it would require to be set as the default file browser to work, which may not be desirable here. If we want a really deep integration I guess we will have to try to detect what file manager it is and explicitly pass the right flags. That means we need a new abstraction for this

Actually we could come up with a new mimetype (`inode/parent-dir`?) and set the new `.desktop` file as the default for that, I think this gives the most control to the user

quadratech188:
Still, I think the simple fix works enough for now

aurelleb:
I agree, let's call it a quick win

PR #911, opened 2025-12-31
fix(wayland): fix dismiss on focus lost not working on wayland

bdsqqq:
hey, really appreciate vicinae, love using it! but not dismissing when I clicked something else was really annoying, so I wrote this fix for myself.

Instead of waiting for a focusloss even lixe in x, we can instead rely on knowinf if we have keyboard focus, from my testing this worked, but this is the first time I ever touched cpp so let me know if there is something I missed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ __cmake*
 /result
 
 CLAUDE.md
+build-dev/
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -5,6 +5,6 @@
 # -DVICINAE_GIT_COMMIT_HASH=XXX (short rev)
 
 release:
-  tag: "v0.17.3"
-  rev: "e8e0c24ce35e1bbfc78029f7ad258f4d073f9e19"
-  short_rev: "e8e0c24ce"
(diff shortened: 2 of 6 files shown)


aurelleb:
I tested your branch and I'm not sure to understand what this fixes, it doesn't change the focus loss behavior for me on Hyprland. If layer shell is enabled with exclusive keyboard interactivity it doesn't close on focus loss (since focus is never lost, I don't think we can do much about this one) and it works as expected when layer shell is disabled or keyboard interactivity is on demand.
I'm assuming this works on your system due to some implementation specific thing, what are you running?

Edit: found a fix for this with #914, this may also solve your issue

bdsqqq:
appreciate you @aurelleb, 
probably a system specific thing, i'm using niri, vicinae, a niri thing to handle gestures, and have a quickshell layer on top of everything for something else, 
will do manual testing removing each layer and testing if the current behavior of main works at some point.

in the worst case, we close and just leave this here for future reference in case anyone has a similar experience

bdsqqq:
ah
just saw your pr, oh well, closing haha

PR #918, opened 2026-01-02
feat: add mkRayCastExtension helper to nix flake outputs

ankarhem:
The[ home manager module](https://github.com/nix-community/home-manager/blob/e4e78a2cbeaddd07ab7238971b16468cc1d14daf/modules/programs/vicinae.nix#L192) contains a helper to declare raycast extensions. This adds that same functionality when using the vicinae flake input directly (as is currently the recommended way in the docs).

To use the helper it is declared like so:

```nix
    extensions =
      [
        (inputs.vicinae.packages.${pkgs.stdenv.hostPlatform.system}.mkRayCastExtension {
          name = "gif-search";
          sha256 = lib.fakeSha256;
          rev = "3ec994afcd05b2b6258b3b71ab8b19d6b6f1e0e4";
        })
      ];
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -64,6 +64,7 @@
       overlays.default = final: prev: {
         vicinae = prev.callPackage ./nix/vicinae.nix { gcc15Stdenv = prev.gcc15Stdenv; };
         mkVicinaeExtension = prev.callPackage ./nix/mkVicinaeExtension.nix { };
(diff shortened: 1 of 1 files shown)


aurelleb:
trying this tool out (free for OSS) not sure I will keep it long term

Issue #930, opened 2026-01-04
Scripts are not detected


michaelschattgen:
**System information**

- Version: v0.18.0 (b22ad4bda)
- Build info: GCC 15.2.0 - Release - LTO
- Provenance: appimage
- OS: openSUSE Tumbleweed - 20251217
- QT Platform: wayland
- DE: Hyprland/Wayland

**Describe the bug**

I saw that #912 got included in its latest release so I was eager to try it out. However after reading through the documentation and through the code it seems like Vicinae does not detect my scripts the way it should.

**To Reproduce**

Create a new script under `~/.local/share/vicinae/scripts/test` with the following contents:
```python
#!/usr/bin/env python3
# @vicinae.schemaVersion 1
# @vicinae.title My Script
# @vicinae.mode fullOutput

print("Hello world!")
```
Make the script executable and run "Reload Script Directories" to make sure it is reloaded.

**Expected behavior**

The script to be available from within the launcher, but nothing happens. I am not entirely sure if this is user error or a bug.

aurelleb:
it could be that this location is not in path due to how your XDG_DATA_DIRS are specified. That said I think I should force `~/.local/share/vicinae/scripts` to be part of the paths that are scanned.
In the meantime you can add this directory manually from the preferences:

<img width="1020" height="313" alt="Image" src="https://github.com/user-attachments/assets/d0508110-e515-48b5-a26b-c48ae6b76d85" />

michaelschattgen:
Prior to creating this issue I went over to the Extensions tab to see if I was able to make adjustments there, but to no avail. The "Script Command" group doesn't show up for me.

<img width="1021" height="623" alt="Image" src="https://github.com/user-attachments/assets/3b2c9772-c798-491c-a49b-43fa5b38a06c" />

aurelleb:
ah that's an interesting one, I think that's because the search only applies to items and not groups, so since there is no script you can't get it unless there is no query entered. That's another issue for me to fix.
In any case I pushed a fix for this and I'm going to make a patch release so your problem should be fixed

michaelschattgen:
The new release works! Thank you for the swift response and version bump.

PR #931, opened 2026-01-04
fix: use common data dirs as default script dirs

aurelleb:
## Summary by CodeRabbit

* **New Features**
  * Improved resource and script discovery mechanism that now automatically searches both user home directories and system-wide standard data directories in a unified manner. This enhancement ensures that application scripts and data files can be reliably discovered from standard system locations, providing better compatibility with system-wide installations and shared resource directories.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


aurelleb:
Fixes #930

PR #978, opened 2026-01-18
feat(api): add support for native node modules

RichardDorian:
This pull request adds support for native node modules. It works by telling ESBuild to load `.node` imports by copying the file to the out directory,  [doc here](https://esbuild.github.io/content-types/#external-file).

Note: ESBuild doesn't support have both `outfile` and `outdir` at the same time. `outfile` was used to specify the name of the output file. However ESBuild doesn't change the name of the file as it is an entry point and the SDK already enforces the name of the file being exactly the command name. Therefore, only defining `outdir` is fine.


aurelleb:
works great, thanks!

PR #979, opened 2026-01-19
mkVicinaeExtension,mkRayCastExtension: use extendMkDerivation

0xda157:
better practice then my previous solution. a breaking change but everything breaking here has been being warned about for a while now.


aurelleb:
thanks!

PR #989, opened 2026-01-21
fix(ui): hide status bar divider when status bar is hidden

Razuer:
## Summary

Fixes a visual artifact where the 1px horizontal divider line above the status bar remained visible even when the status bar itself was hidden.

## The Bug
When `dmenu` is used with `--no-footer` or with a width < 500px (which triggers auto-hide), the status bar content (`m_bar`) is hidden, but the separator line (`m_barDivider`) is not, leaving a floating line at the bottom of the window.

## The Fix
Updated the `statusBarVisiblityChanged` connection in `LauncherWindow` to sync the visibility of `m_barDivider` with `m_bar`.

## Before

image

## After

image


aurelleb:
thank you for your PRs, I have more to say about the other two, I will get back to you later

PR #996, opened 2026-01-22
feat: disable paste on niri

aurelleb:
Until we have a better paste implementation

### DIFF
--- a/vicinae/src/services/window-manager/wayland/wayland.cpp
+++ b/vicinae/src/services/window-manager/wayland/wayland.cpp
@@ -131,7 +131,9 @@ bool WaylandWindowManager::closeWindow(const AbstractWindow &window) const {
   return true;
 }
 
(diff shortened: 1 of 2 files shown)


hced:
I reinstalled my CachyOS/Niri system and discovered that the paste-on-return-key feature was gone within the Clipboard extension.

Would you mind briefly explaining the reason it was removed specifically for Niri?

Is there *any* way I can make my environment support it (again, since it must have 'supported' it prior to reinstalling my system)? I'm not a world class professional Linux user and I don't remember exactly how I made it work before, but I do know that I had `ydotool` installed and maybe there waw something I did with that. I hope it wasn't by adding myself in the "input" group (something I understand carries security implications).

Anyway, I really miss this feature!

aurelleb:
@hced it was too buggy. But I'm working on a proper replacement, as I switched to niri myself! `/dev/uinput`, which is what `ydotool` is using, is indeed something I'm considering since it also has the benefit of being compositor agnositc.

PR #1004, opened 2026-01-23
fix: allow ESC to exit inline argument input focus

Razuer:
## Summary
- Enables ESC key to exit focus from inline argument inputs (used by extensions, scripts, etc.).
- Returns focus to the main search bar instead of closing the application.
- Clears input values on exit to reset the focus/execution flow.

## Details
- **Component**: `InlineQLineEdit` (used in `ArgCompleter`).
- **Fix 1 (UX)**: Intercepts `Qt::Key_Escape` in `InlineQLineEdit` and emits a signal instead of ignoring it (which previously caused the event to bubble up and close the window).
- **Fix 2 (Logic)**: Clears the input values in `ArgCompleter` when ESC is pressed. This ensures that subsequent activation of the item triggers the "focus arguments" logic again, rather than the "execute command" logic (which runs when values are present).

## Usage
1. Open a command with arguments (e.g., an extension or script).
2. Click the command to focus the inline input.
3. Type something.
4. Press ESC -> Focus should return to the main search bar (window stays open).
5. Click the command again -> Input should focus again (instead of executing immediately).

### DIFF
--- a/vicinae/src/ui/arg-completer/arg-completer.cpp
+++ b/vicinae/src/ui/arg-completer/arg-completer.cpp
@@ -54,6 +54,8 @@ void ArgCompleter::setArguments(const ArgumentList &args) {
       emit valueChanged(collect());
     });
 
+    connect(edit, &InlineQLineEdit::escapePressed, this, &ArgCompleter::escapePressed);
+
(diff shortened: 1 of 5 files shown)


aurelleb:
This was fixed some time ago, we switched to QML since

PR #1024, opened 2026-01-27
feat: add DEB packaging support

mookechee:
Add Debian package build infrastructure:
- scripts/packaging/debian/: Debian packaging configuration files
- scripts/mkdeb.sh: Local build script for creating .deb packages
- .github/workflows/build-deb.yaml: CI workflow for DEB builds
- Modify release.yml to include DEB in releases

The package supports Debian 12, Ubuntu 22.04, and Ubuntu 24.04 through compatible dependency declarations.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -5,6 +5,6 @@
 # -DVICINAE_GIT_COMMIT_HASH=XXX (short rev)
 
 release:
-  tag: "v0.19.2"
-  rev: "92cb677fca1513266bc63235e2d01a28f165ee41"
-  short_rev: "92cb677fc"
+  tag: "v0.19.3"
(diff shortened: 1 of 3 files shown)


mookechee:
I originally intended to validate this PR in my forked repository before submitting. However, since the CI workflows rely on self-hosted runners with specific labels (`deb`, `appimage`, `prod`), I'm unable to replicate the actual build environment in my fork.

Therefore, I'm submitting this PR to verify whether my changes work correctly on the main repository's runners. I appreciate your patience and any feedback you may have.

aurelleb:
Hi! Thank you very much for your PR .deb packages are really something that would be cool to have!

That being said, I'm not too sure to understand why we need to build it in the self-hosted runner because the self-hosted runner is definitely not debian. I'm pretty sure the produced binaries won't work on debian.

PR #1043, opened 2026-02-01
feat:  Support for base conversions using libqalculate

### LINKED ISSUE #953: Hexadecimal conversion
No, it's not related to a problem

If possible, adding hexadecimal conversion. Examples could be:
- `172 to hex` == `0xAC`
- `0xAC to number` == `172` - Workaround right now is `0xAC-0`

CopticFelo:
This PR closes #953 
## Summary
This adds support for base conversions (from and to decimal, hexadecimal, binary, octal or roman) using libqalculate

## Details
The function used to calculate expressions is ```CALCULATOR->calculate()``` which doesn't support base conversions so this adds a regex check to use ```CALCULATOR->calculateAndPrint()``` instead when evaluating expressions ending in "to hex" or "-> binary" etc
#### Disclaimer: The regex is generated by GPT 5 but i tested it and it seems to work fine

And also added extra conditions to check if an expression starts with 0x, 0o or 0b and convert it to decimal if so 
## Examples
```
123 to hex = 0x07B
123 to HEXADECIMAL = 0x07B
21 -> roman = XXI
234 -> bin = 0b000000000011101010
0xF8 = 248
0b1110 = 14
```

Please let me know if you want me to change anything and I would be happy to do it.

### DIFF
--- a/src/server/src/extensions/root/root-search-controller.cpp
+++ b/src/server/src/extensions/root/root-search-controller.cpp
@@ -103,7 +103,7 @@ void RootSearchController::startCalculator() {
   bool containsNonAlnum = std::ranges::any_of(m_query, [](QChar ch) { return !ch.isLetterOrNumber(); }) ||
                           m_query.starts_with("0x") || m_query.starts_with("0b") || m_query.starts_with("0o");
   const auto isAllowedLeadingChar = [&](QChar c) {
-    return c == '(' || c == ')' || c.isLetterOrNumber() || c.category() == QChar::Symbol_Currency;
+    return c == '-' || c == '(' || c == ')' || c.isLetterOrNumber() || c.category() == QChar::Symbol_Currency;
   };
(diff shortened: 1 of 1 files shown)


CopticFelo:
Nevermind found a better way to do it without using regex magic, we could just rely on CALCULATOR->calculateAndPrint() for everything and parse the answer and question to get MathStructures, i don't think this breaks anything that already worked before (but should be tested more nonetheless) but it does mean roman numeral conversions don't work anymore as they are treated as unknowns in libqalculate (bug?) and i have not found a way to mark base conversions explicitly as CalculatorAnswerType::CONVERSION without regexes

aurelleb:
if we can do this without regexp it's even better yes, losing roman numeral support is not a pretty big deal I'm pretty sure

aurelleb:
conflicts are my fault, I'm going to fix this on my end don't worry about it

aurelleb:
Tested locally and it works very well, congratulations and thank you!

PR #1075, opened 2026-02-10
fix(docs): Update incorrect settings example

dinckelman:
Nix/Home-Manager has a seemingly copy-pasted value for this, which doesn't actually compile. Update reflects what `vicinae config default` reports


aurelleb:
thanks!

PR #1077, opened 2026-02-10
build(nix): add catch2_3 for debug build

nilsojunior:
This package is needed for compiling the debug build.


aurelleb:
true, thanks! (you could have disabled tests as a workaround, but it's desirable to compile them)

PR #1088, opened 2026-02-13
fix: libqalculate results with repeating decimals not displaying

### LINKED ISSUE #1081: division by float does not work
## **System information**

-

## **Describe the bug**
if you want to divid by the float number (ex. 4.2) the vicinae doest show the solving 

## **To Reproduce**
try to divid something by something (ex. 50/5.5)

## **Expected behavior**
to show the solved math problem

## **Screenshots**

Image

CopticFelo:
This fixes #1081 
## Summary
When evaluating an expression like ```1/3``` (or any expression that would result in a repeating decimal) using the libqalculate backend in Vicinae, the calculator errors out with a message like ```Trailing characters "" (not a valid variable/function/unit) in number "0.333" were ignored.``` referencing the … character (U+2026), the only fix i found so far is just disabling the indicate_infinite_series option in the print options  so instead of ```0.33…``` we would have ```0.33333333```


aurelleb:
thanks!

PR #1095, opened 2026-02-15
Support for algebraic equation in qalculate backend

lufimio:
most of the examples here should work, with some small changes
https://github.com/Qalculate/libqalculate/tree/master?tab=readme-ov-file#examples-expressions
using it held up pretty well to Linear Algebra and Differential Equations homework

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -5,6 +5,6 @@
 # -DVICINAE_GIT_COMMIT_HASH=XXX (short rev)
 
 release:
-  tag: "v0.19.7"
-  rev: "1399606ebe1e799c4c22c7fbcaeae155a6a3d6e4"
-  short_rev: "1399606eb"
+  tag: "v0.19.8"
+  rev: "825cf785cf14fbd0ad253ae72baef6b50c4e74cb"
+  short_rev: "825cf785c"
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -325,6 +325,46 @@ install_desktop_files() {
 	fi
 }
 
+install_browser_manifests() {
+	echo "Installing browser native messaging manifests..." >&2
+
+	local templates_dir="$INSTALL_DIR/usr/share/vicinae/native-messaging-hosts"
+	local native_host_bin="$INSTALL_DIR/usr/libexec/vicinae/vicinae-browser-link"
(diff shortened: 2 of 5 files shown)


aurelleb:
one thing I don't understand is that all these work in qalc out of the box? The fact we have to rely on regexp seems very off to me, I think there should be an easier solution to leverage these capabilities from the library directly. I guess I will go look into what the `qalc` program does because all it does is consume this library as well. Maybe the best solution is going to be invoking qalc every calculation but I'm not the biggest fan of this

lufimio:
The regex is only to fix the issue that happens when solving indefinite integrals, in qalc it doesn't matter if the +C is a unit because it doesn't differentiate between units and non units, but vicinae has the units mode which interprets the +C as coulombs which was the but I wanted to fix. I could do this by checking if the result contains a variable but that would break an equation like "0.05 N/m * x = 4 N" which would be a valid use case in physics.

aurelleb:
> but vicinae has the units mode which interprets the +C as coulombs which was the but I wanted to fix

Can you clarify this? In particular, I'm not sure what you mean by "unit mode". All the calculator has to care about is to compute the expression and display what unit is on the left and what unit is on the right, if any.  if there is no unit we just display generic question/answer text.
Sorry if I don't understand something obvious here, haven't done this kind of math in a long time

Edit: did you mean you can run `qalc -nounits` to disable units or something?

lufimio:
In vicinae, when you do a conversion, "3m/s to ft/min" for example, vicinae runs some extra code to display the full expanded name of the units in the subtext. Qalc doesn't have to worry about that so it can just leave the +C as a unit, but in vicinae, the result of the indefinite integral gets a subtext of coulombs, which shouldn't happen. Handling that edge case is what the regex is for, maybe parsing the mathstructures would be better but it would be just as complex if not more.

> if there is no unit we just display generic question/answer text.

That is the part that's acting weird with indefinite integrals, and what the regexes fix.

aurelleb:
I see yeah. I think it would definitely be more correct to parse the math structs but I agree that the regexp solution is a good and probably simpler workaround. I will check qalc's implementation to see how they handle these cases.
I tested your branch locally and it seems to work fine, although we don't have test suites for the calculator just yet. We should probably add some in the future.

Thanks!

PR #1096, opened 2026-02-15
Fix deeplink documentation spacing in CLI option

bowlofneighs:
The current link in many terminals will open `https://docs.vicinae.com/deeplinks)` which of course returns a 404


aurelleb:
used an alternative phrasing with makes more sense than adding a trailing space, as it would look a bit off

PR #1103, opened 2026-02-19
fix(calculator): normalize operator spacing for SoulverCore backend

damaradiprabowo:
Expressions like '12/6', '10 /2', '10/ 2' failed to compute because SoulverCore requires spaces around operators. Add normalization in calculate() that ensures exactly one space on each side of +, -, *, /, %, and ^ before passing to soulver_evaluate.


aurelleb:
good catch, thanks!

PR #1106, opened 2026-02-20
feat: Setting for exiting on backspace

Carboxide68:
Added a setting that toggles whether a command should be exited on backspace when there's no input. Made default value on, to preserve current behaviour.

### DIFF
--- a/extra/config.jsonc
+++ b/extra/config.jsonc
@@ -18,6 +18,9 @@
 	// Supports "navigate_back" or "close_window"
 	"escape_key_behavior": "navigate_back",
 
+	// navigate back when backspace is pressed on an empty search input
+	"pop_on_backspace": true,
+
 	// If the layer shell protocol is used to position the window and the keyboard interactivity is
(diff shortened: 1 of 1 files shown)


aurelleb:
I'm still okay with that being an option but as I thought, this breaks with the new UI update. Can you fix these conflicts? Also, I think the name of the option should probably be something like "pop_on_backspace" instead, exit is a bit too general.

Carboxide68:
Yes, will fix tomorrow. Is it fine if the setting sits in the general settings section, or should it be placed elsewhere?

aurelleb:
general is fine

Carboxide68:
I have updated the branch, it's working correctly when I've tested it. I might have done something wrong in `launcher-window.hpp` since I'm not providing any way to update the setting dynamically there, but I'm unsure if it even is applicable.

aurelleb:
I can confirm it is working fine. It's reactive by default thanks to the property binding.

PR #1107, opened 2026-02-20
feat(icons): Add Blank icon for invisible spacer

rattletat:
Adds `Icon.Blank` — a transparent SVG that reserves icon space in list items without rendering anything visible.

**Use case:** When a `List` has a mix of items with and without icons, items without icons have their text shifted left, breaking alignment. `Icon.Blank` acts as an invisible spacer to keep text aligned.

**Changes:**
- `src/server/icons/blank.svg` — transparent 32x32 SVG
- `src/typescript/api/src/api/icon.ts` — enum entry


aurelleb:
I'm not a fan of this solution, I think it would be preferable to handle this case explicitly in the API (by treating a `null` icon differently than an `undefined` icon for instance). 
you can still use this trick in your extensions by bundling a blank asset if you want though.

PR #1113, opened 2026-02-23
Niri window manager

refractalize:
This is a window manager implementation for [Niri](https://github.com/niri-wm/niri). We get a few new things vs the generic Wayland window manager:

- windows have workspaces, which show up on the right when using the Switch Window command
- windows are now ordered with most recently focused a the top

Implementation:
- we use the niri socket documented here: https://niri-wm.github.io/niri/IPC.html, so we don't incur any command line startup costs.
- we create two socket connections, one for following the event stream, which allows us to track the window state over time, not unlike how the wayland window manager does it. And secondly, a socket for performing actions, like focus and close window.
- I've intentionally kept the paste support disabled. We can re-enable it but in testing I found it to be a bit flaky so decided to leave it out of this change.

### DIFF
--- a/src/server/src/services/window-manager/niri/niri.cpp
+++ b/src/server/src/services/window-manager/niri/niri.cpp
@@ -239,18 +239,10 @@ bool WindowManager::closeWindow(const AbstractWindow &window) const {
   return true;
 }
 
-bool WindowManager::supportsPaste() const { return m_keyboard.isAvailable(); }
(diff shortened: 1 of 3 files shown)


aurelleb (commented):
very welcome addition, thanks a lot! 
I tested it on niri and it works pretty well.

Paste support was disabled because as you pointed out, it needs further refinement. Hopefully I  can get it to work properly soon.

PR #1114, opened 2026-02-23
feat: migrate frontend to QtQuick

aurelleb:


aurelleb:
@cilginc @ArjixWasTaken @jaredallard
Adds a dependency on [qt6-declarative](https://archlinux.org/packages/extra/x86_64/qt6-declarative/) and [syntax-highlighting](https://archlinux.org/packages/extra/x86_64/syntax-highlighting/) packages. (and their gentoo equivalents).
For all packages (bin & source)

cilginc:
> For all packages (bin & source)

Aren't `-bin` packages statically linked anyway? If that's the case, do we really need to define these dependencies for them?

aurelleb:
@cilginc only problematic libraries like cmark-gfm and protobuf, not fully static. Full static is not desirable anyways

PR #1134, opened 2026-02-28
fix: copy answer when pressing enter on calculator result

lunoxide:
Fixes an issue where pressing Enter on a calculator result executes the wrong action (like searching files).

**Cause:**
When the async calculator result is added to the top of the list, QML re-selects index `0`. The C++ model ignored this request because it thought index `0` was already selected, leaving the action panel configured for the old fallback item.

**Fix:**

* Reset `m_selectedIndex = -1` when async results finish (`handleCalculatorFinished` and `handleFileSearchFinished`) to force the C++ model to rebuild the correct action panel.
* Added a missing `scrollTargetIndex` function to `RootSearchModel` to fix a QML warning in the console.


aurelleb:
good catch, thanks!

PR #1135, opened 2026-02-28
Issue 1111 - Resolution Candidate

### LINKED ISSUE #1111: System Freeze After Triggering Sleep Power Action
**System information**

- Version: v0.19.9 (1a215a715c)
- Provenance: binary_release
- OS: EndeavourOS
- QT Platform: wayland
- DE: KDE/Wayland
- Graphics: Nvidia

**Describe the bug**

After running the sleep command under Power Management, the system immediately goes into a hard system freeze before the kernel can log anything. Recovery requires a hard system shutdown by cutting the power.

Using the command line or other KDE widgets to activate sleep mode works flawlessly. This issue is entirely unique to Vicinae.

**To Reproduce**

- Opening Vicinae
- Type "Sleep"
- Run "Put System to Sleep"
- Confirm the command (if configurations require it)
 
**Expected behavior**

I expected the system to properly go to sleep.

**Screenshots**

N/A

**Additional context**

I have confirmed that the issue doesn't affect GNOME or X11 (on my system).

cypress-exe:
Fix: Graceful Log Out Instead of TerminateUser
    
Calling logind's TerminateUser forcefully kills all processes in them session without giving the compositor a chance to release GPU resources cleanly. On KDE/Wayland with Nvidia, this causes a hard system freeze.

This commit delegates the log out action to the DE's session manager via an async D-Bus call while quitting vicinae immediately so that it doesn't hold resources or block the teardown sequence.

For non-(KDE||Gnome) DEs, falls back to TerminateSession scoped to the current session.

Fixes #1111

Although some of this commit includes code generated from Claude Opus 4.6, I have personally reviewed and tested all parts of this commit.


aurelleb:
I see, that makes sense. Thank you!

PR #1140, opened 2026-02-28
fix: Reimplement considerPreedit

quadratech188:


aurelleb:
thanks, I completely forgot about it :sweat:

PR #1147, opened 2026-02-28
fix: do not reset selection on clipboard history change

### LINKED ISSUE #1145: The selection in the Clipboard History resets to the top when an item is deleted
**System information**

- Version: v0.20.1 (2049d2d39e)
- Provenance: binary_release
- OS: Arch Linux - 
- QT Platform: wayland
- DE: GNOME/Wayland

**Describe the bug**

On the clipboard history, when deleting an entry after the first one, the selection resets to the first item in the list instead of the previous entry (as in the old QWidgets version).

**To Reproduce**

Activate the clipboard history, navigate one or more items below, delete it, the selection resets to the top of the list.

**Expected behavior**

The selection should move to the item below in the list.

**Additional context**

I'm giving a chance to opencode and Opus 4.6 to fix the issue. I hope to have a PR for that ready soon.

aurelleb:
fixes #1145

### DIFF
--- a/src/server/src/qml/clipboard-history-model.cpp
+++ b/src/server/src/qml/clipboard-history-model.cpp
@@ -18,8 +18,10 @@ void ClipboardHistoryModel::setEntries(const PaginatedResponse<ClipboardHistoryE
   if (!m_entries.empty()) {
     sections.push_back({.name = QString(), .count = static_cast<int>(m_entries.size())});
   }
+  bool incremental = !selectFirstOnReset();
   setSections(sections);
   setSelectFirstOnReset(false);
+  if (incremental) refreshActionPanel();
 }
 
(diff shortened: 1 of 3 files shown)


fernandocarletti:
There still a bug if you just avoid resetting the state: it won't delete the next entry. In my fix I had to call the `setSelectedIndex(m_selectedIndex)` on `CommandListModel::setSections()`. I ran your branch and can confirm the bug still there.

This is what I was about to make the PR: https://github.com/fernandocarletti/vicinae/commit/55a6298098494952503358191bd6595d9623b1d3

Just a disclaimer, my C++ and QML is quite rusty since it is more almost 10 years since I last made anything with it, so it was mostly Opus 4.6 getting it.

aurelleb:
yeah I know I have another commit coming. Currently the selection also does not reset on search which is definitely unexpected.

PR #1164, opened 2026-03-05
feat: Made debug mode visible on launcher corner

SiriusCrain:
Added badge on corner to not look into settings to check is prod or debug version is running.
Code generated by `claude-opus-4-6`

image


aurelleb:
I'm not sure how much we need this, it's pretty easy to tell if we are in debug mode from the CLI or the about page in the settings. Besides, if you are using a debug build your probably know it already, because you had to compile it yourself

SiriusCrain:
> I'm not sure how much we need this, it's pretty easy to tell if we are in debug mode from the CLI or the about page in the settings. Besides, if you are using a debug build your probably know it already, because you had to compile it yourself

Yeah, I think it make sense. It is not one-click run of dev mode, so maybe this is kinda redundant. Feel free to close this PR if you think it is too much, do not worry)

aurelleb:
Yeah I think we don't really need this for now, maybe later

PR #1187, opened 2026-03-09
Feat/revamp shortcut keys

pavle99:
- Refactor shortcut rendering so shortcut semantics live in C++ and the UI only renders structured display tokens.
- Apply the new shortcut badge styling everywhere shortcut hints appear, including action panels, footer actions, keybind settings, and recorder previews.
- Update alias badges to match the new visual language while keeping aliases as plain text badges.
- Add server-side shortcut tests and wire the new suite into the project test flow.


aurelleb (changes requested):
I may give it a second pass tomorrow, but I'm submitting this review for the few obvious things I've seen.

I built it locally and I really like the new design, thank you!

PR #1189, opened 2026-03-09
fix: nix devshell to use correct qt environment

schromp:
This moves QT dependencies into a QTEnv.
In the history of nixpkgs this was done internally by qt6.full but that package was removed.
sadly this gives makes us repeat the qt packages from the package definition but oh well :shrug: 

with this devshell i can build and execute vicinae without problems

i also removed nixd and nixfmt because these packages should not be installed by the flake. the user may decide to use another lsp


aurelleb:
thx!

PR #1192, opened 2026-03-09
chore: Update npm deps to mitigate vulnerabilities

SiriusCrain:
Before:
32 vulnerabilities (2 low, 3 moderate, 26 high, 1 critical)

After:
3 vulnerabilities (1 low, 2 high)

### DIFF
--- a/src/typescript/api/package.json
+++ b/src/typescript/api/package.json
@@ -50,7 +50,7 @@
 		"@types/chai": "^4",
 		"@types/mocha": "^10",
 		"chai": "^4",
-		"mocha": "^10",
+		"mocha": "^11.7.5",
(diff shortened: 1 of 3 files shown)


src/typescript/api/package.json
```diff
@@ -31,7 +31,7 @@
 		"@jgoz/esbuild-plugin-typecheck": "^4.0.3",
 		"@oclif/core": "^4",
 		"@oclif/plugin-help": "^6",
-		"@oclif/plugin-plugins": "^5",
+		"@oclif/plugin-plugins": "^2.4.7",
```
aurelleb: why are we downgrading from 3 major versions?

PR #1220, opened 2026-03-16
feat: make font rendering type configurable

iGerman00:
Qt Quick's default distance-field text rendering can appear noticeably thinner, hairier or generally inconsistent on dark backgrounds. This is very noticeable on certain types of displays and looks fairly broken/poor quality. It mostly applies to light text on dark backgrounds, but can also be noticeable against light backgrounds. Users on Linux have system-level fontconfig settings that could provide better rendering results if used and there is an expectation that they will be used wherever possible.

I added a configurable text rendering mode that allows users to choose between two different rendering types:

* Native: Uses platform/fontconfig. Depending on the config - results in thicker, sharper text that respects system font settings
* Qt: Uses Qt's distance-field rendering for potentially better performance with scaling and animations, which I left as the default

Changes:

- Expose text rendering toggle in Advanced Settings page
- add toggle in Advanced Settings UI ("Native font rendering")
- Apply text rendering mode globally on startup and when config change

Added documentation in config templates and example config as well

---

Demo

This requires some explaining. Display types behave very differently on different setups. This demo was taken on a 34" 3440x1440 curved VA display, which does not tolerate poor font rendering at all. For you, it might not be that different, or might look worse, which is why this is an option and not a new default. Demos are difficult to translate between screens, but try to:
1. Open either the video or the GIF (GIF is easier - just open it in a new tab)
2. View in **100% 1:1 scaling** - that means disabling any "Retina" type screen scaling settings and zooming out to 100%
2. Test on different display types and devices 
3. Be considerate to people's preferences if it looks a certain way for you - as it might not look the same to them.

Personally I strongly prefer Windows' RGB subpixel antialiasing font rendering, so I prefer when my new option is enabled.

My demo isn't very good as I needed to cause a rerender (by changing the search query), but the loop will show you the difference pretty clearly.

GIF version (click to open in new tab and view at proper scaling):
vicinae_font_rendering_fix

MP4 version:
https://github.com/user-attachments/assets/44d4c077-1334-4979-a737-3e2dc801f687

---

My font config:
<details><summary>Click to expand</summary>

```xml
<fontconfig>
 <match target="font">
  <edit mode="assign" name="lcdfilter">
   <const>lcddefault</const>
  </edit>
 </match>
 <match target="font">
  <edit mode="assign" name="rgba">
   <const>rgb</const>
  </edit>
 </match>
 <match target="font">
  <edit mode="assign" name="hinting">
   <bool>true</bool>
  </edit>
 </match>
 <match target="font">
  <edit mode="assign" name="hintstyle">
   <const>hintslight</const>
  </edit>
 </match>
 <dir>~/.local/share/fonts</dir>
 <match target="font">
  <edit mode="assign" name="antialias">
   <bool>true</bool>
  </edit>
 </match>
</fontconfig>
```
Plus,
`FREETYPE_PROPERTIES=cff:no-stem-darkening=0 autofitter:no-stem-darkening=0 type1:no-stem-darkening=0 t1cid:no-stem-darkening=0`

</details>

### DIFF
--- a/extra/config.jsonc
+++ b/extra/config.jsonc
@@ -59,6 +59,11 @@
 	"keybinding": "default",
 
 	"font": {
+		// Text rendering backend used by Qt Quick text.
+		// "native": uses platform/fontconfig rendering (can have better dark-mode weight/clarity)
+		// "qt": uses Qt distance-field rendering (typically better performance for heavy scaling/animation)
+		"rendering": "qt",
+
 		// The font family to use for the general vicinae UI.
 		// "auto" (recommended): uses the bundled Inter font
 		// "system": uses the system/platform default font
--- a/src/server/src/cli/server.cpp
+++ b/src/server/src/cli/server.cpp
@@ -46,6 +46,7 @@
 #include "vicinae.hpp"
 #include <filesystem>
 #include <QGuiApplication>
+#include <QQuickWindow>
 #include <csignal>
 #include <QString>
 #include <qlockfile.h>
@@ -56,6 +57,14 @@
 #include "server.hpp"
 
 namespace fs = std::filesystem;
+
+static void applyTextRenderingMode(const config::FontConfig &fontConfig) {
+  if (fontConfig.rendering == "qt") {
+    QQuickWindow::setTextRenderType(QQuickWindow::QtTextRendering);
+  } else {
+    QQuickWindow::setTextRenderType(QQuickWindow::NativeTextRendering);
+  }
+}
 
 void CliServerCommand::setup(CLI::App *app) {
   app->add_flag("--open", m_open, "Open the main window once the server is started");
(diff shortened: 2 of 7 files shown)


aurelleb:
Well, I learned something new today. I can indeed barely tell the difference, and I'm on a 5120x1440 VA curved display myself. Probably my eyes not being sharp enough.
Nothing against adding that as an advanced option, thank you for your contribution and attention to detail!

iGerman00:
If you have some time, check if performance is strongly affected by this. I'm just a drive-by contributor so I don't know much about the codebase and how to properly test for that, but it shouldn't be very noticeable. If it's not meaningful, it would be nice to default to the native option, since on higher PPIs it's hardly noticeable, but it gets more apparent for bigger screens and smaller PPIs; like netbooks, cheaper laptops, or big 1080p TVs.

aurelleb:
To me it looks like it's quite the same, will try running vicinae with it on for a longer period of time and maybe switch to it if it's as fast, we will see.

PR #1230, opened 2026-03-18
build: add amd64 debian package

### LINKED ISSUE #269: Use GitHub action to generate .deb package
**Is your feature request related to a problem? Please describe.**

The current installation method in Debian is to grab the binaries from the release and then place it in `/usr/bin` or `$HOME/.local/bin`. It also involves manually creating a desktop file. It would be nice to have an installable .deb package.

**Describe the solution you'd like**
GitHub action can be utilized to generate a .deb package with the release -- for both x86_64 and arm64

**Describe alternatives you've considered**
Alternative is to download and extract the binaries and place them appropriately, and to place the icon manually and to create a .desktop file.

**Additional context**
AFAIK, it just requires creating a `debian/` folder. After that, github actions should be able to take care of everything. I can give this a try.

jaredallard:
Adds building, and uploading to GH releases, for amd64 debian packages.
I was unable to fully test this because:

* I don't have access to the actions context of this repo
* Self hosted runners are being used, which I cannot mimic
* I built the binary for this package on Gentoo, which doesn't match the
  protobuf versions in Ubuntu, but I was able to get past a lot of
  shared link issues. I likely missed one or two dependencies.

I did not add arm64 support because the self-hosted runners appear to be
amd64 only.

Possible issue with this method: We will rely on the version of libs used in the
runner (e.g., protobuf)

Feel free to take this over the finish line.

Closes: #269


aurelleb:
Well, the main issue with this is not really the packaging aspect of things but rather the really out of date dependencies these systems ship with (debian in particular). We could probably have a working deb package for latest ubuntu but I'm pretty sure the QT version alone would block older systems. Some of the other dependencies are more manageable and can be linked statically like I do for protobuf and cmark-gfm already.

Given what I just said it's likely the appimage + script is the best solution to this issue for older systems, but maybe I'm missing something.

As for arm64, yes the self hosted runner is x86_64 and emulation is too slow to be viable. I could work on getting a github runner ready for that purpose, but they are much slower and require more setup for QT + latest gcc + node etc...

jaredallard:
Don't hate me immediately, but maybe in that case we could ship an app image inside? It's not the most orthodox but it would solve that problem

aurelleb:
> Don't hate me immediately, but maybe in that case we could ship an app image inside? It's not the most orthodox but it would solve that problem

I'm not sure whether it's really orthodox or not but if it works it works, I don't really care, and I don't see a better solution. What you could try to do, is look at what I do in the script install (which uses the appimage) and try to tweak it for deb packages. That's probably the best approach.

Jelloeater:
@aurelleb @jaredallard Would be really nice to have this. Having issues installing my desktop, works fine on my laptop.

OS: Ubuntu MATE noble 24.04 x86_64
Host: Lenovo ThinkServer TS130 (110568U)
Kernel: Linux 6.8.0-101-generic
Packages: 3667 (dpkg), 2 (flatpak-user), 21 (snap), 74 (brew)
Shell: zsh 5.9
DE: Mate 1.26.2
WM: Marco (X11)
WM Theme: Colloid-Dark
Theme: Colloid-Dark [GTK2/3/4]
Icons: Tela-circle-dark [GTK2/3/4]
Font: Noto Sans (12pt) [GTK2/3/4]
Cursor: macOS-BigSur (24px)
Terminal: guake
CPU: Intel(R) Xeon(R) E3-1225 V2 (4) @ 3.60 GHz
GPU: NVIDIA GeForce GTX 1060 3GB [Discrete]
Locale: en_US.UTF-8

PR #1232, opened 2026-03-18
fix: correctly wait for the custom locker to be done executing

tretre91:
When specifying a custom locker in the power management options, it would get killed 30 seconds after locking the session.
The `QProcess` used to launch the custom locker is destroyed (and kills the underlying program) after the call to `waitForFinished()`, which only waits 30s by default. Using `waitForFinished(-1)` fixes the issue.

Running a debug build of vicinae gives this warning (`loquet` being the custom locker):
```
[18:20:47.638] WARN   -  QProcess: Destroyed while process ("loquet") is still running.
```


aurelleb:
Interesting, for some reason this didn't trigger with hyprlock. good catch nonetheless thanks.

Issue #1235, opened 2026-03-19
Option to launch apps in new systemd scopes


BBaoVanC:
## **Is your feature request related to a problem? Please describe.**

Apps running in a session under systemd are supposed to have their own unit (meaning separate cgroup as well). This makes things like resource usage tracking and OOM killing work better. Instead, currently vicinae just spawns programs and they remain under the `vicinae.service` unit.

## **Describe the solution you'd like**

An option that allows spawning new systemd scopes for apps that get launched through vicinae.

This would also obsolete the `KillMode=process` line in the systemd unit: https://github.com/vicinaehq/vicinae/blob/bbaeb53cf4e017f5255d4d4aa5536f79b699303e/extra/vicinae.service#L14. This line could be removed if this feature is implemented. In my opinion this is the more correct way and `KillMode=process` is just a band-aid.

## **Additional context**

<img width="898" height="293" alt="Image" src="https://github.com/user-attachments/assets/4998d75f-2472-4a6f-8a78-a660edaaf2aa" />

Alternatively, see how niri spawns scopes for anything spawned with `spawn` or `spawn-sh`:

<img width="693" height="1442" alt="Image" src="https://github.com/user-attachments/assets/691ce787-e1c7-4aee-9d17-14a9e03feef6" />

They say wlr-which-key because that's the little keyboard shortcut tool I'm using to spawn them. At some point in the future I might change it to use `systemd-run` instead so the units get better names.

BBaoVanC:
Actually this is really weird. I just updated Vicinae from 0.20.4 to 0.20.7 and now it automatically launches with uwsm. I am not even using uwsm but I have it installed. It's possible that this issue should be closed and the solution is to just install uwsm.

aurelleb:
It tries to auto detect uwsm usage at startup and uses it

aurelleb:
Also, did you try using systemd-run as an app launcher prefix?

<img width="883" height="562" alt="Image" src="https://github.com/user-attachments/assets/e1f400ce-a16e-4863-8212-b6c612be7656" />

pavle99:
@aurelleb If I prefix it with `systemd-run` (and pass `--user` flag to avoid the password prompt) the app opens and immediately closes

BBaoVanC:
> Also, did you try using systemd-run as an app launcher prefix?

[...]

PR #1238, opened 2026-03-19
feat: Option to launch apps in new systemd scopes

SiriusCrain:
Fixes https://github.com/vicinaehq/vicinae/issues/1235

Used most straightforward path, yes no bin name since pid is not given during arg set, but easier than write logic myself rather than rely on QProcess


aurelleb:
app prefix does this already, as discussed externally

PR #1255, opened 2026-03-23
chore: Update dockerfile

SiriusCrain:
node:22 is in maintanance mode, recommended to switch to 24.
Ubuntu can be updated as well

### DIFF
--- a/scripts/runners/appimage/AppImageBuilder.Dockerfile
+++ b/scripts/runners/appimage/AppImageBuilder.Dockerfile
@@ -5,7 +5,7 @@
 # Note: In order for linuxqtdeploy to work, the docker environment needs to be passed the fuse device.
 # Typically done like so: docker run --cap-add SYS_ADMIN --device /dev/fuse <image_name>
 
-FROM ubuntu:24.10 as base-builder
+FROM ubuntu:22.04 as base-builder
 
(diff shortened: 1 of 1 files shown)


aurelleb:
not that simple. this WILL break many things. Unless you built the image yourself and made sure everything is working. Btw the old image is by design: https://docs.vicinae.com/build-appimage

SiriusCrain:
> not that simple. this WILL break many things. Unless you built the image yourself and made sure everything is working. Btw the old image is by design: https://docs.vicinae.com/build-appimage

Sure, if ubuntu version can cause troubles we can restore it. Node and QT should not, I suppose

Edit: new docker image is not building locally, and I do not want to fix it so reverted ubuntu version for now

aurelleb:
you don't know until you try, if you change this stuff you need to make sure it still builds the appimage correctly (the CI won't pick up on your changes, the appimage runner is prebuilt once and pulled by the runner).

SiriusCrain:
> you don't know until you try, if you change this stuff you need to make sure it still builds the appimage correctly (the CI won't pick up on your changes, the appimage runner is prebuilt once and pulled by the runner).

Will check ubuntu fixes tommorow, or maybe even in another PR since it can be breaking change

SiriusCrain:
No need for now

PR #1258, opened 2026-03-25
[fix] Force cancel to focus onAboutToShow

### LINKED ISSUE #1257: Alerts will focus Cancel after the open animation is done
**System information**

- Version: v0.20.7 (6d9fa2883a)
- Provenance: binary_release
- OS: Arch Linux - 
- QT Platform: wayland
- DE: Hyprland/Wayland

**Describe the bug**

When performing an action that requires a confirmation, such as power off system, vicinae will ask for a confirmation with an alert dialog. While the dialog is opening, the action already listened for keyboard input and it's possible to press Tab to already select confirm. However, once the animation is over, the cancel option is selected again, making is hard to quickly pressed Enter, Tab, Enter to run the action.

It's worth noting that if done slow enough, it's not a problem. It done really fast, it's possible to confirm before the alert opening animation is done. But if done at the right speed, it will prevent me from choosing confirm.

**To Reproduce**

- Highlight Power Off System
- Press Enter, then as fast as possible, Tab
- While the dialog is opening, confirm is selected
- Once the animation is done, cancel is selected

**Expected behavior**

It should default to cancel before the dialog opens, not after. Side note, it would also be nice to have a way to disable the animation entirely.

**Screenshots**

N/A

**Additional context**

N/A

CapitaineToinon:
Move the forceActiveFocus call to onAboutToShow instead of onOpened to prevent focus to change after the alert has been opened.

Fixes #1257

Note: I tried running make format but it formatted around 30 files unrelated to the PR. Is this known? Am I doing something wrong?

### DIFF
--- a/src/server/src/qml/qml/AlertDialog.qml
+++ b/src/server/src/qml/qml/AlertDialog.qml
@@ -13,8 +13,10 @@ Popup {
 
     property bool _confirmed: false
 
-    onAboutToShow: _confirmed = false
-    onOpened: Qt.callLater(cancelBtn.forceActiveFocus)
+	onAboutToShow: {
+		_confirmed = false;
+		Qt.callLater(cancelBtn.forceActiveFocus);
+	}
     onClosed: {
         if (!_confirmed)
             launcher.alertModel.cancel();


aurelleb:
Thank you for the fix!

> Note: I tried running make format but it formatted around 30 files unrelated to the PR. Is this known? Am I doing something wrong?

Interesting, qml or C++ files? Might have to look into that.

CapitaineToinon:
It was QML files but bevermind, I had qmlformat installed from qt5 in my PATH instead of qt6. With the correct qmlformat it's all good now, although the format isn't correct in my PR for the AlertDialog.qml file. I guess a single formatting error can be left for a future PR.

aurelleb:
yeah it's alright the CI doesn't enforce qml formatting, only C++ for now

PR #1259, opened 2026-03-25
feat: Add link preview

### LINKED ISSUE #1256: Support embed preview in clipboard
Is it possible that instead of link text it shows embed in preview too? Like i copy links and generally forget what were they about and you render the icons anyways is it possible we can get embed previews as well?

SiriusCrain:
Closes https://github.com/vicinaehq/vicinae/issues/1256

slopped with claude

image

Updated design:

image

### DIFF
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -10,14 +10,14 @@ if (USE_SYSTEM_KF6)
 	find_package(KF6SyntaxHighlighting REQUIRED)
 endif()
 
-list(APPEND LIBS  
+list(APPEND LIBS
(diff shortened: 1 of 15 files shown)


bpavuk (commented):
I don't have expertise to evaluate most of C++ code, but some of it does rub me the wrong way, aside from C++ being biblically verbose in general

SiriusCrain:
Sorry for delay, will fix in a week (on vacation right now)

SiriusCrain:
@aurelleb updated

vMohammad24:
if this is getting merged, make it optional/opt-out, I don't think its very secure to fetch EVERY link you copy.

PR #1262, opened 2026-03-26
feat: Add text selection and context menu to markdown view

iGerman00:
When using an extension I noticed it was impossible to copy text since both Ctrl+C and right click paths are not disabled for MarkdownView, unless there is a copy button on a code block, yet it was possible to select text, which is fairly confusing from a UX standpoint. This PR enables keyboard shortcuts and adds a right-click context menu for the view.

Context menu uses the ActionItemDelegate component for visual consistency. Screenshots below with two different themes, it should work nicely and look great with all themes.

Extension used for demo: [ChatGPT (Raycast)](https://www.raycast.com/abielzulio/chatgpt)

Screenshot_20260326_233529

Screenshot_20260326_233548


aurelleb:
very nice thanks!

PR #1271, opened 2026-03-29
fix(form): fixes vertical padding for forms

FredrikMWold:
This pr fixes an issue with the vertical padding in forms.
Currently there is only top padding not bottom padding.

Before:
https://github.com/user-attachments/assets/4a93a95b-eb9b-4f67-89f7-dc28c02f04ec

After:
https://github.com/user-attachments/assets/96ba9163-2e65-49fc-bfe9-0a64d637a1af


aurelleb:
good catch, thanks!

PR #1272, opened 2026-03-29
feat: replace protobuf with figura (custom codegen)

aurelleb:
figura is a tiny codegen tool purpose built for vicinae

it directly implements the service layer with client/server code generation for both C++ and TypeScript (required by extensions). 

json-rpc is used as a communication protocol (request-response + one way notifications)

For our use case, figura is a replacement for protobuf (and also covers grpc stuff)

- Speed should remain similar (serialization/deserialization is not the bottleneck anyways)
- Easier to debug
- Way simpler to maintain and modify
- Less compilation overhead

### DIFF
--- a/nix/vicinae.nix
+++ b/nix/vicinae.nix
@@ -1,5 +1,4 @@
 {
-  abseil-cpp,
   cmake,
   cmark-gfm,
   fetchNpmDeps,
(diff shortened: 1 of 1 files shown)


aurelleb:
@cilginc @ArjixWasTaken @jaredallard 
starting with v0.20.10 protobuf/abseil are no longer required, they can be removed from the dependencies. abseil is probably not listed explicitly as it is a transitive dependency required by protobuf

PR #1273, opened 2026-03-29
fix: fixes issue where holding backspace triggers popOnBackspace

### LINKED ISSUE #1117: Holding backspace to erase the search bar triggers "escape_key_behavior"
When I hold down backspace to erase everything in the search field it also triggers the "escape_key_behavior"  behavior as soon as the field is empty. 

That triggering the "escape_key_behavior" behavior  would require single press of the backspace key.

Have not found any alternatives.

FredrikMWold:
This pull request proposes a small change to the backspace handling logic in the `SearchBar.qml` component. I have found it annoying that holding the backspace to erase the search input almost always ends up with me on the roots search. 

Now, pressing and holding the backspace key will not repeatedly trigger the "go back" action only a single, non-repeated backspace press will do so.

Fixes #1117

### DIFF
--- a/src/server/src/qml/qml/SearchBar.qml
+++ b/src/server/src/qml/qml/SearchBar.qml
@@ -290,7 +290,7 @@ Item {
                         event.accepted = true;
                     } else if (_handleNavigation(event)) {
                         event.accepted = true;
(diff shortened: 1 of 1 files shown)


aurelleb:
makes sense, thx

PR #1274, opened 2026-03-30
fix: changes labels for multiline fields from vertical-center to top-left aligned

FredrikMWold:
This pr changes the position of form labels for multi line inputs.

The way it is now with the label vertically centered make the reading flow of the form a little of. Your eyes have dart around to gather all the context of the field. 

This changes them top top-left aligned instead

image
image


aurelleb:
thx!

PR #1288, opened 2026-04-05
fix: update tsconfig.json boilerplate to be typescript v6 compatible

FredrikMWold:
The latest release of VS Code updates the built-in TypeScript compiler to v6. This introduces a breaking change: Node.js types are no longer automatically included. As a result, projects now need to explicitly specify `"types": ["node"]` in their `tsconfig.json`.

More details: https://gist.github.com/privatenumber/3d2e80da28f84ee30b77d53e1693378f

This PR adds the required `types` field to all `tsconfig.json` files in the project, including those in the extension boilerplate.

The change is backward compatible, so projects still using TypeScript v5 should not experience any issues.

### DIFF
--- a/extra/extension-boilerplate/tsconfig.json
+++ b/extra/extension-boilerplate/tsconfig.json
@@ -12,6 +12,7 @@
 		"skipLibCheck": true,
 		"forceConsistentCasingInFileNames": true,
 		"resolveJsonModule": true,
-		"jsx": "react-jsx"
+		"jsx": "react-jsx",
+		"types": ["node"]
 	}
 }
--- a/src/lib/emoji/scripts/emoji/tsconfig.json
+++ b/src/lib/emoji/scripts/emoji/tsconfig.json
@@ -104,6 +104,7 @@
 
     /* Completeness */
     // "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
-    "skipLibCheck": true                                 /* Skip type checking all .d.ts files. */
+    "skipLibCheck": true,                                /* Skip type checking all .d.ts files. */
+    "types": ["node"]
   }
 }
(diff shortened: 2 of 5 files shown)


FredrikMWold:
I think it might be a good idea to have an llm create a small codemod to add it to all existing extensions

aurelleb:
afaik they were never auto included? I mean maybe vscode did that, but I always had to add @types/node for it to work with tsserver LSP

FredrikMWold:
image

That was based on this in there migration guide.
If i don't add this line i get and error on every import for node, but if i downgrade to typescript v5 the errors go away.
So i think they where auto importing something before.

aurelleb:
oh I see, doesn't that also cause issues for other types like for react?

FredrikMWold:
Though it would too but its not complaining about anything.

The docs says that most people should just have `"types": ["node"]`, but "types": ["*"] would mirror the same behavior as v5
image
https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/

FredrikMWold:
Found a good explanation for why node needs to be included, but not react:
https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#types-now-defaults-to-[]

aurelleb:
well I read this and it doesn't seem to explain it? I don't understand why we don't need to load the react types since the default behavior now requires to explicitly specify the types to import?

FredrikMWold:
As I understand it, when you import something from React, TypeScript automatically loads the corresponding types from `node_modules/@types/react`.

However, with node, you typically import modules like `fs` rather than something explicitly named node. Because of this, typescript doesn’t automatically include the node type definitions.

In typescript v6, if you want it to load types from `node_modules/@types/node`, you need to explicitly add "node" to the types array in your `tsconfig.json`, since there’s no direct import that triggers it.

Before v6 typescript just loaded all type definition from `node_modules/@types/*` even if they are never used

aurelleb:
ah yeah for all these global types you never import explicitly it kinda makes sense

PR #1291, opened 2026-04-06
fix: the issue-template referred to deprecated command to report bugs

imrehg:
The old shortcut usually was just not doing anything silently. One has to check the vicinae server logs to see that the `extensions` ways of calling things is deprecated, while still not having a clue of how to update the link without further guidace).

The updated link correctly opens the issue template from the command line.


aurelleb:
actually it has been broken for even longer, because I forgot to update the syntax of the old one which changed a little in the past. Thanks for fixing it!

PR #1295, opened 2026-04-07
feat: option to keep pinned when clearing clipboard

### LINKED ISSUE #1276: [Feature Request] "Clear Clipboard History" should preserve Pinned items
Currently, the "Clear Clipboard History" command (and the "Delete all" action) performs a full wipe of the clipboard database. While the warning "All your history will be lost" is technically accurate, it is somewhat counter-intuitive that Pinned entries are not excluded from this bulk deletion.

As someone who uses Pinned entries to store persistent templates (like FAQs and installation guides), I found that a routine history cleanup inadvertently removes these critical "static" snippets, requiring a manual database restoration.

I suggest that the "Clear History" function should treat pinned entries as protected by default.

Coming from the Gnome Clipboard History extension, I am used to a workflow where "Favorites" (or Pinned items) are preserved even after a system reboot or a manual "Clear All" command. I would like Vicinae to adopt a similar behavior:

- When a user clicks "Remove all" or "Clear History", only the unpinned items should be deleted.
- Pinned items should remain saved as a "persistent layer" of the history until manually unpinned or deleted individually.

I am currently performing "surgical" cleanups via terminal to keep my pinned items intact:
sqlite3 clipboard.db "DELETE FROM selection WHERE pinned_at IS NULL;"

Alternatively, this behavior could be configurable through a setting such as:

"Preserve pinned items when clearing history"

andersmmg:
Closes #1276

### DIFF
--- a/src/server/src/extensions/clipboard/clipboard-extension.cpp
+++ b/src/server/src/extensions/clipboard/clipboard-extension.cpp
@@ -6,6 +6,8 @@
 #include "extensions/clipboard/clipboard-history-command.hpp"
 #include "service-registry.hpp"
 
+bool ClipboardExtension::preservePinnedOnClear = false;
+
 class ClipboardClearCommand : public BuiltinCallbackCommand {
   QString id() const override { return "clear"; }
   QString name() const override { return "Clear Current Clipboard Data"; }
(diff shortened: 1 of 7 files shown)


FredrikMWold:
I tested this and it froze my computer and turned it into a yet engine for a few seconds 😆

Probably because i had like 40k entries and the loop that has to delete every single one individually now instead of just deleting the entire directory.

Maybe a solution could be a seperate data dir 'm_pinnedDataDir` Then store and move the entries when something is pinned or unpinned? This would allow for simpler cleanup

SiriusCrain:
@andersmmg I can investigate and fix perfomance issue if it's ok for you, and push it here

PR #1312, opened 2026-04-17
fix: close launcher before executing action to avoid race condition

### LINKED ISSUE #1047: Active application windows not focused
**System information**

- Version: v0.19.3 (4233336d)
- Provenance: local
- OS: Fedora Linux Asahi Remix 42 (Forty Two [Adams]) - 42
- QT Platform: wayland
- DE: Hyprland/Wayland

---

On hyprland, when attempting to use vicinae to focus a window using `shift+enter` or via the "Switch Windows" command, the window is not focused.

What does (sometimes) happen is the workspace the window is on is made active, but the window itself is not focused. Sometimes a completely random workspace is made active (with no discernible correlation to the desired window).

It does work for some windows. In my case, the _only_ window I have found to work is `ghostty`. 

**To Reproduce**

In my case:

1. Be on Hyprland
2. Install chromium (for example)
3. Launch chromium
4. Open/focus another window/workspace
5. Use the "Switch Windows" command, find the open chromium window, press enter

**Expected behavior**

I expect the workspace to be activated, and the selected window in the workspace to be focused. To be explicit, I want the behaviour to be identical to:

```bash
hyprctl dispatch focuswindow address:0xaaaab51abae0
```

Which operates perfectly as described.

strangeoceans:
Resolves #1047 

I was running into the same issue using `river-classic`. In my case, I was able to successfully swap focus via `vicinae app launch <id>`, but when I did it via the launcher window, the launcher window would close and focus would return to the same window it was on before opening launcher.

As @aurelleb suspected, this seems to have been caused by some sort of race condition in the WM related to closing the launcher window immediately after executing the switch window action. Moving this line such that the launcher window is closed *before* executing the action fixes the issue for me.

### DIFF
--- a/src/server/src/navigation-controller.cpp
+++ b/src/server/src/navigation-controller.cpp
@@ -447,10 +447,10 @@ void NavigationController::executeAction(AbstractAction *action) {
     }
   }
 
+  if (action->autoClose()) { closeWindow({.clearRootSearch = true}); }
+
   action->execute(&m_ctx);
   closeActionPanel();
-
-  if (action->autoClose()) { closeWindow({.clearRootSearch = true}); }
 }
 
 AbstractAction *NavigationController::findBoundAction(const QKeyEvent *event) const {


aurelleb:
ah, thanks a lot!

Issue #1329, opened 2026-04-23
Window switcher doesn't scroll viewport to focused window on Hyprland scrolling layou


justindotdevv:
**System information**

- Version: v0.20.12 (8401bab2cf)
- Build info: GCC 15.2.1 - Release - LTO
- Provenance: binary_release
- OS: Arch Linux - 
- QT Platform: wayland
- DE: Hyprland/Wayland

**Describe the bug**

When using the window switcher to focus a window on Hyprland with the scrolling layout enabled, the compositor switches to the correct workspace and gives the target window keyboard focus, but the viewport does not scroll to reveal the target column. The user is left staring at whatever columns were previously visible, with no indication of where the focused window actually is.

This works correctly on the dwindle layout because all tiled windows are already visible on-screen — focus alone is sufficient

**To Reproduce**

Set Hyprland layout to scrolling (general:layout = scrolling)
Open several windows so that some columns are off-screen
Open vicinae's window switcher
Select a window that is in an off-screen column (or on another workspace)
Observe: workspace switches correctly, but the viewport does not scroll to the focused column

**Expected behavior**

After selecting a window, the scrolling layout viewport should scroll to bring the target column into view.

**Screenshots**

**Additional context**

Root cause
---

`focusWindowSync` in `hyprland.cpp` dispatches only:

```cpp
dispatch focuswindow address:0x...
```

The `focuswindow` dispatcher sets compositor-level focus but does not trigger the scrolling layout's viewport adjustment. The scrolling layout has a dedicated `layoutmsg fit active` command that scrolls the viewport to show the currently focused column, but vicinae never sends it.

Proposed fix
---

Append `dispatch layoutmsg fit active` as a batch command alongside `focuswindow`:

```cpp
[[BATCH]]dispatch focuswindow address:0x...;dispatch layoutmsg fit active
```

`layoutmsg fit active` is a no-op on layouts that don't support it (dwindle, master), so this is safe to send unconditionally. The `[[BATCH]]` prefix ensures both commands execute atomically in a single compositor tick with no extra socket connection.

aurelleb:
I see nothing wrong with that fix but is there any reason for focuswindow to not do that automatically?

justindotdevv:
focuswindow in Hyprland updates focus and optionally switches workspaces, but never notifies the active layout to adjust its viewport.

PR #1330, opened 2026-04-23
fix: scroll viewport to focused window on Hyprland scrolling layout

justindotdevv:
**fixes** https://github.com/vicinaehq/vicinae/issues/1329

On Hyprland's scrolling layout, the window switcher focuses the target window but doesn't scroll the viewport to reveal its column. This is because focuswindow only updates compositor-level focus without notifying the layout to adjust the viewport.

Fix: Batch dispatch layoutmsg fit active alongside focuswindow. This tells the scrolling layout to scroll the viewport to the focused column. It's a no-op on other layouts (dwindle, master)

### DIFF
--- a/src/server/src/actions/wm/window-actions.hpp
+++ b/src/server/src/actions/wm/window-actions.hpp
@@ -8,15 +8,17 @@ class FocusWindowAction : public AbstractAction {
   std::shared_ptr<AbstractWindowManager::AbstractWindow> m_window;
 
   void execute(ApplicationContext *ctx) override {
-    auto wm = ctx->services->windowManager();
-    wm->provider()->focusWindowSync(*m_window.get());
+    auto window = m_window;
+    ctx->navigation->deferUntilWindowHidden([ctx, window]() {
+      auto wm = ctx->services->windowManager();
+      wm->provider()->focusWindowSync(*window.get());
+    });
+    ctx->navigation->closeWindow({.clearRootSearch = true});
   }
 
 public:
   FocusWindowAction(const std::shared_ptr<AbstractWindowManager::AbstractWindow> &window)
-      : AbstractAction("Focus window", ImageURL::builtin("app-window")), m_window(window) {
-    setAutoClose();
-  }
+      : AbstractAction("Focus window", ImageURL::builtin("app-window")), m_window(window) {}
 };
 
 class CloseWindowAction : public AbstractAction {
--- a/src/server/src/navigation-controller.cpp
+++ b/src/server/src/navigation-controller.cpp
@@ -34,6 +34,16 @@ bool NavigationController::hasCompleter() const {
   return false;
 }
 
+void NavigationController::deferUntilWindowHidden(std::function<void()> fn) {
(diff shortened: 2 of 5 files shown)


aurelleb:
any reason this is in draft state? looks good to me?

justindotdevv:
doing testing on my machine to ensure it is working as intended

justindotdevv:
Root cause was action timing rather than just the Hyprland IPC command.

Vicinae was calling `focusWindowSync()` immediately when the action fired, before the launcher had actually hidden. On Hyprland, the launcher often runs as a layer-shell surface with exclusive keyboard interactivity, so on same-workspace switches Hyprland was still dealing with Vicinae’s layer shell when the focus request arrived. In practice that meant the cursor warped to the target window, but the target did not reliably become the effective focused window soon enough for scrolling follow-focus to kick in.

That also explains why cross-workspace looked better: `dispatch focuswindow` changes workspace first in that case, and that larger workspace transition masked the sequencing problem.

Why previous iterations failed:
  - `fit active` moved the viewport but incorrectly resized the target column.
  - `center` in the same batch still ran too early to be reliable.
  - even the later “focus first, then center after polling” workaround was still happening before Vicinae had fully hidden.

Final fix:
  - defer `focusWindowSync()` until after the launcher has actually hidden
  - then run the Hyprland focus request
  - keep the scrolling-specific `layoutmsg center` follow-up in the Hyprland backend for same-workspace scrolling

With that sequencing fixed, same-workspace switching now behaves correctly on Hyprland scrolling without the `fit active` resize side effect.

aurelleb:
this workaround seems very convoluted to me.

It looks like 9fb1f6d2f882ebf36ab19919e99ca36ad7e06c9b fixed this for some people? I'm assuming this didn't fix it for you?

justindotdevv:
> It looks like https://github.com/vicinaehq/vicinae/commit/9fb1f6d2f882ebf36ab19919e99ca36ad7e06c9b fixed this for some people? I'm assuming this didn't fix it for you?

yea it did not fix it for me, this pr currently fixes it for me. this fix does touch a lot more files than i expected. after many iterations this is what i can provide unfortunately.

justindotdevv:
out of curiosity i pulled the latest commit and tested it. it worked for me. traced the fix to 9ce527e. exclusive keyboard interactivity was preventing the compositor from cleanly handing off focus, so the scrolling layout's follow_focus never triggered. downgrading to OnDemand lets the focus transfer work properly.

aurelleb:
oh okay, interesting. I guess that could be considered an Hyprland bug no? If so I guess we can close this and maybe document the workaround?

justindotdevv:
agreed, this is a hyprland-side limitation. 9ce527e already fixes this so this PR is no longer needed.

Documenting for future reference:
- Symptom: scrolling layout doesn't scroll viewport to the focused window after switching via vicinae.
- Cause: keyboard_interactivity: exclusive on the layer-shell surface blocks timely focus handoff.
- Fix: use a build with 9ce527e (OnDemand interactivity). No user config .

PR #1352, opened 2026-05-06
fix: Navigation with keyboard now scroll view

SiriusCrain:
Previously if keyboard focused item goes outside of visible view it was not following it.


aurelleb:
ah yeah  saw this one happening many times, thx!

PR #1364, opened 2026-05-09
fix: cap arg completer width to enable text auto-scroll

### LINKED ISSUE #1116: [FR] Shortcut input box max-width limit and auto-scroll to end
I have a Shortcut (quicklink) for Perplexity. I use it tens of times a day.
Sometimes I need longer prompts that overflow the full width of the query input box.

Right now, when the text is longer than Vicinae width, the text at the end is cut off and not visible.

Auto-scroll to the right of the input box.

## **Additional context**
When it's short, it's fine:
Image

When the query is long, the end is not visible:
Image

oxysoft:
## Summary

- Inline command argument text fields (`ArgCompleter.qml`) grew unboundedly with text content via `implicitWidth: textMetrics.advanceWidth + 16`, preventing QML's native `TextInput` horizontal auto-scroll from ever engaging
- Caps each arg delegate's `implicitWidth` to a proportional share of available width (minus icon and spacing), floored at 60px
- The existing `clip: true` on `TextInput` then triggers auto-scroll naturally once text exceeds the bounded container
- Same fix applied to dropdown delegates for consistency

Closes #1116

## Test plan

- [ ] Open a command with text argument(s), type a long string — text should now scroll horizontally instead of overflowing
- [ ] Short arguments should still grow-to-fit naturally (no visual change)
- [ ] Commands with multiple arguments should split available width evenly
- [ ] Dropdown arguments should also remain bounded

### DIFF
--- a/src/server/src/qml/qml/ArgCompleter.qml
+++ b/src/server/src/qml/qml/ArgCompleter.qml
@@ -60,6 +60,10 @@ RowLayout {
             required property var modelData
 
             readonly property bool isLast: index === root.visibleArgs.length - 1
+            readonly property real maxArgWidth: {
+                var totalSpacing = root.spacing * (root.visibleArgs.length + 1);
+                return Math.max((root.width - 25 - totalSpacing) / root.visibleArgs.length, 60);
+            }
 
             Layout.alignment: Qt.AlignVCenter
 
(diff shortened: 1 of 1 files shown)


aurelleb:
very nice fix, thanks!

PR #1373, opened 2026-05-12
nix support for snippets

vMohammad24:
creates a nix wrapper for vicinae-input-server so we can give it cap_dac_override+ep without having any problems

if any nix users want to test this before it gets merged:

change vicinae.url to `github:vMohammad24/vicinae/nix-input-server`

add `inputs.vicinae.nixosModules.default` to your flake input modules

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -78,5 +78,15 @@
         mkRayCastExtension = prev.callPackage ./nix/mkRayCastExtension.nix { };
       };
       homeManagerModules.default = import ./nix/module.nix self;
+
+      nixosModules.default = {pkgs, ...}: {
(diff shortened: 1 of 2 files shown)


aurelleb:
oh wait, I'm not sure what this entails exactly but do we need to depend on your module? or can we make it part of the vicinae repository? 
I mean, so that we do not depend on your github serving what is needed at github:vMohammad24/vicinae/nix-input-server

vMohammad24:
nonono this is just to test before this gets merged, after this is merged you only need to do the 2nd step (vicinae.url stays the same as the current one `github:vicinaehq/vicinae`)

PR #1375, opened 2026-05-13
fix: hyprland lua IPC support

### LINKED ISSUE #1372: Switch Window command broken on Hyprland 0.55
**System information**

- Version: v0.20.15 (a0b875caf)
- Provenance: binary_release
- OS: EndeavourOS - 
- QT Platform: wayland
- DE: Hyprland/Wayland

**Describe the bug**
The "Switch Windows Window Management" command does not work anymore on Hyprland since Hyprland 0.55 and the introduction of the new Lua config

**To Reproduce**

Use the "Switch Windows" command and try to switch to any open application

**Expected behavior**

The workspace and focus should switch to the selected application

aurelleb:
We use both the old and new call in a [BATCH] which actually works, since unknown commands are ignored.

fixes #1372


jemabaris:
Can confirm that it works again. Thank you for handling it so fast! It's a feature I've really learned to love 😊

aurelleb:
you are welcome!

PR #1378, opened 2026-05-13
feat: client side drop shadows

aurelleb:
draws drop shadows on the client side, only if we are a layer surface. We assume that non layer surfaces may already have drop shadows drawn by the compositor.

In order for this to work together with blur, the compositor must implement [ext-background-effect](https://wayland.app/protocols/ext-background-effect-v1) protocol. For compatibility reasons we don't allow drop shadows if the compositor doesn't implement it, at least for now.

### DIFF
--- a/extra/config.jsonc
+++ b/extra/config.jsonc
@@ -106,7 +106,8 @@
 		"client_side_decorations": {
 			"enabled": true,
 			"rounding": 10,
-			"border_width": 1
+			"border_width": 1,
+			"shadow_size": 16
 		},
 
 		// In compact mode, vicinae only shows a search bar in the root search if no query is entered, only expanding to its full size when searching.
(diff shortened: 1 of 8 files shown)


iGerman00:
This broke the interaction with the Glass KWin effect for me @aurelleb 
Killing client side decorations works as a way to fix the ugly blur border, but it also kills rounded corners. 
Would appreciate if there was an option to toggle specifically this new behavior.


Screenshot_20260516_014517

aurelleb:
is there a reason you are not using the builtin ext-background-effect implementation? kde blur is not customizable enough I guess?
You can go in your config and set the shadowSize to 0, it should work. It's not in the json schema but you can check with `vicinae config default`

iGerman00:
> is there a reason you are not using the builtin ext-background-effect implementation? kde blur is not customizable enough I guess?
> 
> You can go in your config and set the shadowSize to 0, it should work. It's not in the json schema but you can check with `vicinae config default`

Liquid Glass style blurring is the reason :P, I helped code it, gotta rep it.

People tend to use these kinds of effects to control blur from a centralized spot with per-class granularity. I do too. 

Will try changing the config tomorrow.

iGerman00:
Sorry, took me a bit longer. That does seem to be working.

Screenshot_20260517_140924

PR #1386, opened 2026-05-14
Vacuum file-indexer after rebuild to reclaim disk space

Heyian:
## Summary

After reducing search paths and running `Rebuild File Index`, `file-indexer.db` and `file-indexer.db-wal` stayed the same size (or grew). The rebuild deletes all rows but never tells SQLite to hand the freed space back to disk, so the file size only ever reflects the largest the index has ever been.

## Fix

After the bulk delete, run `VACUUM` and a truncating WAL checkpoint on the writer thread, then start the new scans. The compact happens while the database is empty, so it's fast. Errors are logged and don't block the rebuild.

## Testing

- Built and ran locally.
- Indexed a broad set of paths, noted the file sizes.
- Reduced the paths, hit `Rebuild File Index`.
- Both files shrank to match the new content.

## clang-tidy

`make check-format` is clean, and `clang-tidy` reports no new issues from this PR. Two pre-existing `bugprone-throwing-static-initialization` errors on `main` (`SQLITE_PRAGMAS` and `EXCLUDED_PATHS`) are left as-is.

## AI disclosure

The code was reviewed but not written by Claude Opus 4.7 using Claude Code.
The PR text was also drafted by it.

### DIFF
--- a/src/server/src/services/files-service/file-indexer/db-writer.cpp
+++ b/src/server/src/services/files-service/file-indexer/db-writer.cpp
@@ -73,6 +73,13 @@ void DbWriter::deleteAllIndexedFiles(std::function<void()> onComplete) {
   });
 }
 
+void DbWriter::compact(std::function<void()> onComplete) {
+  submit([onComplete = std::move(onComplete)](FileIndexerDatabase &db) {
+    db.compact();
+    if (onComplete) { onComplete(); }
+  });
+}
+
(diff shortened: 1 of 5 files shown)


aurelleb:
Thanks. Note that we will probably replace the entire file search stack very soon, not announcing anything yet but it's definitely coming. In the meantime, that's a nice fix.

PR #1405, opened 2026-05-17
feat: fuzzy rerank file search results

pavle99:
- Fetches a wider SQLite FTS candidate pool for file search
- Makes every query word prefix-searchable, not just the last one
- Reranks candidate basenames with Vicinae’s existing fzf matcher
- Supports split fuzzy queries like fol exa finding example folder
- Applies pagination after fuzzy reranking
- Adds regression tests for fuzzy file-search behavior
- Updates TypeScript API docs for file-search behavior


src/server/tests/file-indexer-query-engine-tests.cpp
```diff
@@ -0,0 +1,118 @@
+#include <catch2/catch_test_macros.hpp>
```
aurelleb: you can drop the tests, they are not particularly useful and linking to QT just for that is a mess (and it's a temporary fix anyways)

src/server/CMakeLists.txt
```diff
@@ -896,10 +896,18 @@ if (BUILD_TESTS)
 	find_package(Catch2 3 REQUIRED)
 	add_executable(${TEST_TARGET}
 		tests/shortcut-tests.cpp
+		tests/file-indexer-query-engine-tests.cpp
```
aurelleb: you can drop this as well

aurelleb (changes requested):

pavle99:
@aurelleb Done, dropped the query-engine tests and removed the extra CMake test links.

PR #1416, opened 2026-05-19
feat(theme): added LibAdwaita themes

rflrkn:
Heyyo! I created a Vicinae theme based on the LibAdwaita colour palette, so that Vicinae can blend in a bit better for GNOME users. :)

Vicinae theme switcher

Vicinae About window

Vicinae About window


aurelleb (changes requested):
I think using the blue accent color as as a selection color is way too strong and cause obvious design problems.  The selection should be a subtle but noticeable highlight on the selected item. And, in general, not use an accent color.

image


image

rflrkn:
Valid points! I took a look at some official Gnome apps and tweaked the colours accordingly.

PR #1420, opened 2026-05-19
fix: update theme icons for light and dark variants

ovehbe:
Fix theme icons so they display correctly in both light and dark modes.


aurelleb:
thx!

PR #1423, opened 2026-05-20
fix(install): install application icon to system icon directory

ovehbe:
## Summary

- The install script extracted the AppImage icon into the internal lib tree but never copied it to `$PREFIX/share/icons/`, so desktop environments could not resolve `Icon=vicinae` from the `.desktop` file
- Add `install_icons()` to copy the hicolor icon and refresh the icon cache
- Add icon cleanup to the uninstall path

## Test plan

- [x] Verified icon is installed to `/usr/local/share/icons/hicolor/512x512/apps/vicinae.png` after running `sudo bash install.sh`
- [x] Desktop environment picks up the icon for the `.desktop` entry

Made with [Cursor](https://cursor.com)

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -341,21 +341,21 @@ install_desktop_files() {
 install_icons() {
 	echo "Installing application icons..." >&2
 
-	local icons_source="$INSTALL_DIR/usr/share/icons"
-	local icons_dest="$PREFIX/share/icons"
+	local icon_source="$INSTALL_DIR/usr/share/icons/hicolor/512x512/apps/vicinae.png"
+	local icon_dest_dir="$PREFIX/share/icons/hicolor/512x512/apps"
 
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -338,6 +338,27 @@ install_desktop_files() {
 	fi
 }
 
+install_icons() {
+	echo "Installing application icons..." >&2
+
+	local icons_source="$INSTALL_DIR/usr/share/icons"
```
aurelleb: you install all icons in /usr/share/icons but later you only remove vicinae.png on uninstall
if you uninstall one icon specifically then you should also mirror that during installation, only install that specific icon so that if we ever add some, we do not end up with leftover icons after uninstall

aurelleb (changes requested):

PR #1426, opened 2026-05-20
feat: fff for file search

dmtrKovalenko:
This is how it works:

https://github.com/user-attachments/assets/b9627617-b274-4a38-9ca3-13d6e7c8e963

A few important descisions/questions:

1. fff doesn't persist the index, the index is extremely fast so I don't see a reason to vaste resources storing paths anywhere. The index lives per the scope of open window triggered by file search view open
2. fff uses databases for access patterns, I suppose a lot of people would be happy if we reuse their existing database, I added a logic for this. Though your recently accessed files feature is separate
3. The search open window now shows the amount of indexed files in real time, can revert this - just thinking this is cool


src/server/CMakeLists.txt
```diff
@@ -903,3 +896,41 @@ if (BUILD_TESTS)
 	target_link_libraries(${TEST_TARGET} PRIVATE Catch2::Catch2WithMain Qt6::Core Qt6::Gui)
 	target_compile_features(${TEST_TARGET} PUBLIC cxx_std_23)
 endif()
+
+# Standalone smoke test for the fff integration. Not gated on BUILD_TESTS,
```
dmtrKovalenko, line 984: I've been using this for my local development, it is LLM generated, so keep your own judgement - I think before merge we should get rid of this

dmtrKovalenko:
On the current build? Hmm we should canonicalize the path probably

vMohammad24:
did some testing with it, and i mainly only have the following issues:

1. it indexes everytime i open the menu
2. it does not work with `~` and does not provide any errors (outside of logs) and simply says no results
3. reopening it breaks it entirely (as shown in the video)

https://nest.rip/f/rcBEr (video's too large for github 10MB limit)

PR #1434, opened 2026-05-23
feat: Expose app actions from the footer logo

pavle99:
- Make the bottom-left Vicinae footer logo clickable on the root search view.
- Open the existing action panel from the logo with app-level actions: settings, shortcuts, store, docs, bug report, about, and quit.
- The action panel can now align left for this footer menu while preserving the normal right-aligned Actions button behavior.

### DIFF
--- a/src/server/src/extensions/vicinae/report-bug-command.hpp
+++ b/src/server/src/extensions/vicinae/report-bug-command.hpp
@@ -1,40 +1,6 @@
 #pragma once
+#include "bug-report-url.hpp"
 #include "builtin-url-command.hpp"
-#include <QGuiApplication>
-#include <qurlquery.h>
-#include "environment.hpp"
-#include "lib/os-release.hpp"
-#include "generated/version.h"
-
-static const QString ISSUE_TEMPLATE = R"(**System information**
-
-- Version: %1 (%2)
-- Build info: %3
-- Provenance: %4
-- OS: %5
-- QT Platform: %6
-- DE: %7
-
-**Describe the bug**
-
-A clear and concise description of what the bug is.
-
-**To Reproduce**
-
-Steps to reproduce the behavior.
-
-**Expected behavior**
-
-A clear and concise description of what you expected to happen.
-
-**Screenshots**
-
-If applicable, add screenshots to help explain your problem.
-
-**Additional context**
-
-Add any other context about the problem here.
-)";
 
 class ReportVicinaeBugCommand : public BuiltinUrlCommand {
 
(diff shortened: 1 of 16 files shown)


src/server/src/qml/action-panel-controller.cpp
```diff
@@ -11,6 +12,7 @@ ActionPanelController::ActionPanelController(ApplicationContext &ctx, QObject *p
     : QObject(parent), m_ctx(ctx) {}
 
 ActionPanelView *ActionPanelController::activeRoot() const {
+  if (m_transientRoot) return m_transientRoot;
```
aurelleb: we shouldn't have to use this "transient root" thing for this. We should just set a different action panel when clicking on the root search icon and then restore the current action panel that we had before. The action controller logic shouldn't have to change for this.

src/server/src/qml/action-panel-controller.cpp
```diff
@@ -63,6 +65,27 @@ void ActionPanelController::open() {
   if (m_open) return;
   if (!m_hasActions) return;
 
+  if (m_alignLeft) {
```
aurelleb: same thing here, I don't see why we need alignment concerns in the action panel controller as this is mostly a presentation concern

src/server/src/extensions/vicinae/report-bug-command.hpp
```diff
@@ -36,6 +36,29 @@ If applicable, add screenshots to help explain your problem.
 Add any other context about the problem here.
 )";
 
+inline QUrl makeVicinaeBugReportUrl(const QString &title = {}) {
```
aurelleb: we should put this utility in a separate file, as we shouldn't have to import the extension definition itself to reuse this behavior.

src/server/src/qml/launcher-window.cpp
```diff
@@ -539,6 +544,59 @@ QRect LauncherWindow::cursorScreenGeometry() const {
...
+        ctx->navigation->showHud(QStringLiteral("Opened in browser"));
+      }));
+  helpSection->addAction(new StaticAction(QStringLiteral("About Vicinae"), ImageURL::builtin("info-01"),
+                                          [](ApplicationContext *ctx) {
+                                            ctx->navigation->closeWindow();
+                                            ctx->settings->openTab(QStringLiteral("about"));
+                                          }));
+
+  auto *quitSection = state->createSection();
+  auto *quit = new StaticAction(QStringLiteral("Quit Vicinae"), ImageURL::builtin("power"),
```
aurelleb: not sure about this one, depending on the way vicinae is run it may yield to unexpected behavior. Maybe just remove it for now.

aurelleb (changes requested):
I like the idea but I think the implementation is not good, all we need to implement this is:
- different action panel placement
- swap action panels and restore the one for the current item when its closed. Maybe a simpler solution would be to have two action panel UI elements, one for the current, contextual set of actions and the other for the root search actions. The overhead of using two would be negligible so maybe that's a good solution.

Also other comment, make sure that the panel can be toggled when clicking multiple times on the button, currently it just reopens it.

pavle99:
@aurelleb Thanks, that makes sense. I updated the implementation to keep the action panel controller generic: no transient root and no alignment state there anymore.

aurelleb:
took this as an opportunity to improve the architecture. We now use a separate panel for the general vicinae menu

PR #1452, opened 2026-05-28
chore: remove rang, clean dead code, push_back → emplace_back

dacrab:
### Remove `rang` dependency
`rang` is an unmaintained header-only terminal color library (last commit 2019). It was only used in 3 places in `cli.cpp` for red error output and a reset on exit. Replaced with plain ANSI escape codes (`\033[31m` / `\033[0m`), removing the vendored file entirely.

### Dead code removal (`src/cli/src/theme.cpp`)
- `CheckThemeCommand::run` contained a large commented-out block referencing `rang` — removed
- `TemplateThemeCommand` had an unused `std::optional<std::filesystem::path> m_path` member — removed

### `push_back` → `emplace_back`
Mechanical replacement across the codebase per the coding guidelines. `emplace_back` constructs in-place and avoids an unnecessary copy when inserting values. Cases using `{...}` aggregate init or already using `std::move` were left as-is (they're already optimal).


aurelleb:
no

PR #1455, opened 2026-05-29
fix(xdg): handle footclient in inferTermExec

### LINKED ISSUE #1417: xdg-terminals.list doesn't use the app-id specified in script commands at least for foot
**System information**

- Version: v0.20.15 (a0b875caf)
- Provenance: nix
- OS: NixOS 26.05 (Yarara) - 26.05
- QT Platform: wayland
- DE: niri/Wayland

**Describe the bug**

If there is a xdg-terminals.list file defined and the content is:
```
footclient.desktop
```

The app-id specified in a script-command won't be used.

**To Reproduce**

See above.

**Expected behavior**

The terminal should have the app-id.

Keyruu:
## Problem

When `xdg-terminals.list` points at `footclient.desktop`, the `appId` (and `title`/`workingDirectory`/`hold`) set on a script command is silently dropped.

## Root cause

`XdgAppDatabase::inferTermExec()` matches on `app.program()` (the `Exec=` binary). The `footclient.desktop` shipped with foot has `Exec=footclient`, so `app.program()` returns `"footclient"`, the `"foot"` branch is skipped, and the fallback `{.exec = "-e"}` is used. `launchTerminalCommand()` then omits the corresponding flags because they're absent in the `TerminalExec` mapping.

## Fix

`footclient` accepts the same `--app-id` / `--title` / `--working-directory` / `--hold` flags as `foot`, so reuse the same mapping for both program names.

Fixes #1417

### DIFF
--- a/src/server/src/services/app-service/xdg/xdg-app-database.cpp
+++ b/src/server/src/services/app-service/xdg/xdg-app-database.cpp
@@ -379,7 +379,7 @@ xdgpp::DesktopEntry::TerminalExec XdgAppDatabase::inferTermExec(const XdgApplica
         .hold = "--hold",
     };
   }
-  if (app.program() == "foot") {
+  if (app.program() == "foot" || app.program() == "footclient") {
     return {
         .exec = "-e",
         .appId = "--app-id",


aurelleb:
I guess `footclient.desktop` should also have the xdg-terminal-exec keys then, but the fix is fine.

PR #1457, opened 2026-05-29
feat: add custom keyword for emojis

aurelleb:


src/server/src/qml/emoji-grid-model.cpp
```diff
@@ -81,6 +83,32 @@ class ResetEmojiSkinToneAction : public AbstractAction {
...
+public:
+  void execute(ApplicationContext *ctx) override {
+    auto emojiService = ctx->services->emojiService();
+    auto view = new EditKeywordsViewHost(
+        [emojiService, emoji = m_emoji]() {
+          return QString::fromStdString(emojiService->mapMetadata(emoji).keyword);
+        },
+        [emojiService, emoji = m_emoji](const QString &kw) {
+          return emojiService->setKeywords(emoji, kw.toStdString());
+        },
```
ArjixWasTaken, line 96: This how you do lambdas in C++?

aurelleb: yes lol, with a capture group in this case.

PR #1464, opened 2026-05-31
feat: add various unicode symbols in emoji picker

### LINKED ISSUE #1342: Add other chars to Emoji Picker (just like macOS and Windows versions of Raycast)
I miss the ability to use **Emoji Picker** to get other special chars, like I'm used to do in **Raycast** _(on both Windows and macOS)_.

For example, when I need an arrow _(→)_ or an en-dash _(–)_, if I'm using Raycast I just summon Emoji Picker and use them from there _(they're pinned along with my favorite emojis)_.

But in **Vicinae's Emoji Picker** I get only emojis. That's OK, but it would be great to get the same feature I have in Raycast. 😅

Thanks in advance. Cheers.
Cris

aurelleb:
This PR adds support for various unicode symbols in the emoji picker,  renamed 'Search emojis & Symbols'.
We support most of the useful unicode ranges while trying to avoid useless repetition of symbols

Emoji search is now multi-threaded in order to account for the substantial increase in processable entries.

Fixes #1342

image


aurelleb:
Fixes #1342

PR #1467, opened 2026-06-01
feat(root-search): optionally hide empty-state suggestions/favorites

mmclinton:
## Motivation

The launcher's empty state always shows Suggestions and Favorites. I prefer a minimal launcher when nothing has been typed, and I suspect I'm not alone, so this adds two opt-in toggles to make that possible. I've kept this brief on behavioral design choices; happy to discuss them and the alternatives I considered.

## What's added

**Settings → General** gets two new toggles:

- **Hide suggestions** — drops the frecency-ranked Suggestions list when the search bar is empty.
- **Hide favorites** — drops the Favorites section when the search bar is empty.

Both default to `false`. Zero behavior change for existing users.

When both flags hide everything the empty state would show (no suggestions and no favorites/news to fall back on), the window collapses to just the search bar and footer. A footer **Expand** button — and `↓` on the search bar — temporarily restores the full layout for the current visibility cycle without flipping the toggles. Other arrow keys and `Enter` are intentionally no-ops while collapsed.

## Implementation

- New `rootSearch.hideSuggestionsWhenEmpty` / `hideFavoritesWhenEmpty` flags in `config.hpp` with a matching `Partial<RootSearch>` for the user-config patch path. Defaults preserved.
- `RootSearchModel::rerunSearch()` gates the relevant `setItems()` calls on the flags and on a transient `userExpanded` override.
- `LauncherWindow` tracks `userExpanded` (reset on each visibility cycle) and exposes `userExpand()` to the Footer button and the search-bar `Keys.onDownPressed` handler.
- Drive-by: gives `WindowCompactMode::enabled` an explicit `= false` default to match surrounding convention. No runtime change; closes a latent uninitialized-bool path.

## Manual verification

Branch is rebased on `main` (`67290dff`). Verified locally on Linux/Wayland:

- Defaults: visual parity with `main`.
- Each toggle independently / both together → expected sections appear/disappear immediately on toggle.
- Settings persist across restarts (checked `settings.json`).
- Both toggles on + no favorites/news → window collapses; `↓` and footer **Expand** restore full layout; Up/Left/Right/Enter are no-ops while collapsed.
- Typing a query restores the full layout regardless of toggle state.

## Screenshots

vicinae-collapsed

*Collapsed launcher bar — the new optional default empty-state.*

<br>

vicinae-settings

*Settings pane, General settings, new toggles.*

<br>

**Happy to add a short screen recording of the Expand interaction on request.**

### DIFF
--- a/src/server/src/config/config.hpp
+++ b/src/server/src/config/config.hpp
@@ -181,6 +181,16 @@ template <> struct Partial<TelemetryConfig> {
   std::optional<bool> systemInfo;
 };
 
+struct RootSearch {
+  bool hideSuggestionsWhenEmpty = false;
+  bool hideFavoritesWhenEmpty = false;
+};
+
+template <> struct Partial<RootSearch> {
+  std::optional<bool> hideSuggestionsWhenEmpty;
+  std::optional<bool> hideFavoritesWhenEmpty;
+};
+
 using KeybindMap = std::map<std::string, std::string>;
 
 using ProviderMap = std::map<std::string, ProviderData>;
@@ -214,6 +224,7 @@ struct ConfigValue {
   FontConfig font;
   ThemeConfig theme;
   TelemetryConfig telemetry;
+  RootSearch rootSearch;
 
   WindowConfig launcherWindow;
   Header header;
@@ -265,6 +276,7 @@ template <> struct Partial<ConfigValue> {
   std::optional<Partial<FontConfig>> font;
   std::optional<Partial<ThemeConfig>> theme;
   std::optional<Partial<TelemetryConfig>> telemetry;
+  std::optional<Partial<RootSearch>> rootSearch;
 
   std::optional<Partial<WindowConfig>> launcherWindow;
   std::optional<Partial<Header>> header;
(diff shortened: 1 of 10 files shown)


mmclinton:
Fixed in `a44a0447`. Heads-up for the record: `make format` silently skips the C++ `clang-format` step when `biome` isn't on `PATH`. The `tsfmt` prerequisite fails first and short-circuits the rest of the target. `biome` isn't in the nix dev shell, so I'd missed that. Happy to look into a small Makefile patch in a separate PR, after this has been addressed.

aurelleb:
I think you were looking for compact mode, which does exactly what you want but better, you only have the search bar. You can only enable it from the config file directly for now, I'm going to fix that! 
image

I don't think this feature does anything better than compact mode, and the granularity doesn't earn us much.

mmclinton:
Thanks for the response. You're right, `compact_mode` covers the core use case. Closing in favor of your existing implementation.

If it's useful, I'd be happy to take a stab at the GUI toggle you mentioned wanting to add (something under General → Compact Mode seems like a minimal visual change). Otherwise, no obligation, feel free to ignore.

aurelleb:
feel free to, just make sure you are respecting the existing pattern for it.

PR #1480, opened 2026-06-05
Enable local currency conversion in Qalculate backend

### LINKED ISSUE #1479: Qalculate backend should use locale/default currency for bare currency amounts
## Problem

The Qalculate backend supports explicit currency conversion syntax such as:

```text
1500 USD to EUR
```

but it does not appear to preserve qalc/libqalculate's useful behavior where a bare foreign-currency amount is converted to the user's local/default currency.

For example, in `qalc` a user can type:

```text
1500usd
```

and get a conversion to their local/default currency, while explicit targets still work normally:

```text
1500usd to eur
```

## Expected behavior

For a bare currency amount with no explicit target currency, the Qalculate backend should use the local/default currency selected by libqalculate/qalc. Ideally this should follow the user's locale by default, unless an explicit default currency has been configured.

Examples:

```text
1500usd        -> convert USD to the user's local/default currency
1500usd to eur -> convert USD to EUR explicitly
```

## Local qalc behavior

On my machine, qalc has:

```ini
default_currency=ARS
ignore_locale=0
local_currency_conversion=1
```

With that config:

```text
qalc -t '1500usd'        -> ARS 2150854.497
qalc -t '1500usd to eur' -> €1288.659794
qalc -t '1500usd to ars' -> ARS 2155496.611
```

With a clean qalc config under my current locale (`LANG=en_IE.UTF-8`), bare USD converts to EUR, so qalc/libqalculate can derive the implicit target from locale when no explicit default currency is configured.

## Related but distinct issues

This is distinct from exchange-rate refresh issues such as #743 / #983, and distinct from natural-language phrasing issues like #1199 (`in` vs `to`). This request is specifically about preserving qalc/libqalculate's implicit target currency behavior for bare currency amounts.

repparw:
## What changed

Enable `local_currency_conversion` in the Qalculate backend evaluation options.

## Why

Vicinae already supports explicit currency conversions such as `1500usd to eur`, but bare currency amounts like `1500usd` do not produce the local/default currency conversion that qalc/libqalculate can provide.

With this option enabled, libqalculate can convert bare foreign-currency amounts to the user's local/default currency while preserving explicit target conversions.

## Validation

- `git diff --check`
- Standalone libqalculate probe in `nix develop`:
  - old behavior with `POST_CONVERSION_BEST` and `local_currency_conversion = false`: `$1500`
  - new behavior with `POST_CONVERSION_BEST` and `local_currency_conversion = true`: `EUR 1288.6598`
- Root CMake configure was attempted and reached `libqalculate! calculator backend is enabled`, but could not complete from this sparse checkout because unrelated directories such as `src/lib/*`, `vendor/*`, and `src/typescript` were not checked out.

Fixes #1479.

### DIFF
--- a/src/server/src/services/calculator-service/qalculate/qalculate-backend.cpp
+++ b/src/server/src/services/calculator-service/qalculate/qalculate-backend.cpp
@@ -170,6 +170,7 @@ bool QalculateBackend::supportsCurrencyConversion() const { return true; }
 
 void QalculateBackend::initializeCalculator() {
   m_evalOpts.auto_post_conversion = POST_CONVERSION_BEST;
+  m_evalOpts.local_currency_conversion = true;
   m_evalOpts.structuring = STRUCTURING_SIMPLIFY;
   m_evalOpts.parse_options.parsing_mode = PARSING_MODE_ADAPTIVE;
   m_evalOpts.parse_options.units_enabled = true;


aurelleb:
is this ready to merge or actually a draft?

repparw:
actually a draft. will change it when I get it working

PR #1482, opened 2026-06-06
feat: x11 global shortcut support

aurelleb:
Pure X11 global shortcut support (no XWayland)


aurelleb:
Tested as working on my xfce VM

PR #1484, opened 2026-06-06
fix: input server failing to start on NixOS

Nyramu:
Hello. I noticed I couldn't choose to paste the clipboard content as the clipboard history default action.
After investigating the issue I found out the input server failed to start every time because `VICINAE_INPUT_SERVER_BIN` was set to `/run/wrappers/bin/vicinae-input-server`, which can't be found because it's located in the nix store.
It now points to the executable in the nix store, and it seems to work just perfect now

I encountered this problem and solved it while being on nixpkgs unstable 26.11

Errors found via `journalctl`:
`[V] 2026-06-06T09:54:57 error - input server process error occured "Child process set up failed: execve: No such file or directory" `
`[V] 2026-06-06T09:54:57 error - Failed to start input server QProcess::FailedToStart`

### DIFF
--- a/nix/vicinae.nix
+++ b/nix/vicinae.nix
@@ -66,7 +66,7 @@ gcc15Stdenv.mkDerivation (finalAttrs: {
     cmark-gfm
     kdePackages.layer-shell-qt
     kdePackages.qtkeychain
-	kdePackages.qtshadertools
+    kdePackages.qtshadertools
     kdePackages.syntax-highlighting
     libqalculate
     nodejs
@@ -85,6 +85,11 @@ gcc15Stdenv.mkDerivation (finalAttrs: {
     npmRoot=src/typescript/extension-manager npmDeps=${finalAttrs.extensionManagerDeps} npmConfigHook
   '';
 
+  postFixup = ''
+    wrapProgram "$out/bin/vicinae" \
+      --set VICINAE_INPUT_SERVER_BIN "$out/libexec/vicinae/vicinae-input-server"
+  '';
+
   qtWrapperArgs = [
     "--prefix PATH : ${
       lib.makeBinPath [
@@ -92,7 +97,6 @@ gcc15Stdenv.mkDerivation (finalAttrs: {
         (placeholder "out")
       ]
     }"
-    "--set VICINAE_INPUT_SERVER_BIN /run/wrappers/bin/vicinae-input-server"
   ];
 
   meta = {


aurelleb:
@vMohammad24 can you comment on that?

vMohammad24:
this doesn't use the wrapper at all, it would requrie the user to give permissions manually using setcap, you're having this issue because you did not add the nixos module to your flake outputs, the docs were updated for this change you can find them at [here](https://docs.vicinae.com/nixos#:~:text=vicinae%2EnixosModules%2Edefault)

Nyramu:
> this doesn't use the wrapper at all, it would requrie the user to give permissions manually using setcap

I didn't have to set anything manually nor via my NixOS configuration, if you mean the change I made in this pr

> you're having this issue because you did not add the nixos module to your flake outputs, the docs were updated for this change you can find them at [here](https://docs.vicinae.com/nixos#:~:text=vicinae%2EnixosModules%2Edefault)

Unfortunately doing that does not fix the problem for me. I followed every step in the docs but it goes back to that very same error I provided in the initial comment. Is there anything I can check for?

I forgot to mention it, but I'm on nixpkgs unstable 26.11 if that helps

Edit: I saw the nixos module and as far as I know the source is something under nix/store once you expand the `${self.packages.${pkgs.stdenv.hostPlatform.system}.default}` expression. I could be wrong though

vMohammad24:
this pr changes the input server to the default one (in libexec), as you can see below, the wrapper has correct permissions whereas libexec executable does not. 

from what i can see in your [flake.nix](https://github.com/Nyramu/.dotfiles/blob/main/flake.nix) you do not register the module meaning the wrapper never gets created meaning you can not use the input-server. you do need to add this module, you can see an example [here](https://github.com/vMohammad24/nixos-config/blob/cbb4539e46db58c2c27ab9e9c1b3d4f3fedab35d/flake.nix#L63)
```
➜ getcap /run/wrappers/bin/vicinae-input-server
/run/wrappers/bin/vicinae-input-server cap_dac_override,cap_setpcap=ep
~ 
➜ getcap /nix/store/qi59ziifncsfc2rzq67mq3bk9f3b3bf6-vicinae-0.21.5/libexec/vicinae/vicinae-input-server
~ 
➜
```

Nyramu:
> this pr changes the input server to the default one (in libexec), as you can see below, the wrapper has correct permissions whereas libexec executable does not.
> 
> from what i can see in your [flake.nix](https://github.com/Nyramu/.dotfiles/blob/main/flake.nix) you do not register the module meaning the wrapper never gets created meaning you can not use the input-server. you do need to add this module, you can see an example [here](https://github.com/vMohammad24/nixos-config/blob/cbb4539e46db58c2c27ab9e9c1b3d4f3fedab35d/flake.nix#L63)
> 
> ```
> ➜ getcap /run/wrappers/bin/vicinae-input-server
> /run/wrappers/bin/vicinae-input-server cap_dac_override,cap_setpcap=ep
> ~ 
> ➜ getcap /nix/store/qi59ziifncsfc2rzq67mq3bk9f3b3bf6-vicinae-0.21.5/libexec/vicinae/vicinae-input-server
> ~ 
> ➜
> ```

Yes, currently my repo is out of sync, I was making some local changes.

After your first suggestion, I tried the module, but it didn't change anything.

I tried again because I wanted to send what I saw, and I realized I'd set it up incorrectly before, so it works now.

Thanks for your help. 😃

vMohammad24:
you're welcome, please close this if you don't need anything else.

PR #1488, opened 2026-06-07
feat: Add footer status command menu

pavle99:
- Fix searchable dropdown positioning so compact dropdowns anchor correctly.
- Make the footer status area clickable for active commands, not only the root menu.
- Add a lean command menu with preferences, alias, favorite, deeplink, and extension-specific actions.


pavle99:
Superseded by the split PRs: #1489 for the dropdown positioning fix and #1490 for the footer status command menu.

PR #1489, opened 2026-06-07
fix: anchor native popup windows below their trigger on Wayland

pavle99:
QtWayland positions Popup.Window popups exclusively through an xdg_positioner - the QML x/y are never part of it, and a plain Popup gets the default top-right anchor on its trigger rect instead of the below-the-control anchoring Qt's own ComboBox/Menu/ToolTip opt into. 

This adds a PopupPlacement attached type that sets the positioner override properties QtWayland reads at popup-surface creation, and applies it to the three native popups (searchable dropdowns, tooltips, shortcut recorder). No compositor detection; native popups and their blur/shadow are preserved.

### DIFF
--- a/src/server/src/qml/qml/SearchableDropdown.qml
+++ b/src/server/src/qml/qml/SearchableDropdown.qml
@@ -32,6 +32,10 @@ Item {
         completionPopup.open();
     }
 
+    function popupX() {
+        return Math.min(0, triggerButton.width - completionPopup.width);
+    }
+
     Keys.onReturnPressed: {
         if (!completionPopup.visible)
             open();
@@ -96,7 +100,10 @@ Item {
         id: completionPopup
         parent: triggerButton
         popupType: Popup.Window
-        x: compact ? triggerButton.width - width : 0
+        // On Wayland the compositor places the native popup window from the
+        // PopupPlacement anchor; x/y only apply on other platforms.
+        PopupPlacement.alignment: root.compact ? Qt.AlignRight : Qt.AlignLeft
+        x: root.popupX()
         y: triggerButton.height + 4
         width: Math.max(compact ? 200 : 250, root.width)
         focus: true
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -936,6 +936,8 @@ set(VICINAE_QML_SOURCES
 	src/qml/markdown/text-selection-controller.hpp
 	src/qml/background-effect-attached.hpp
 	src/qml/background-effect-attached.cpp
+	src/qml/popup-placement-attached.hpp
+	src/qml/popup-placement-attached.cpp
 	src/qml/shortcut-inhibitor-attached.hpp
 	src/qml/shortcut-inhibitor-attached.cpp
 )
--- a/src/server/src/qml/popup-placement-attached.cpp
+++ b/src/server/src/qml/popup-placement-attached.cpp
@@ -67,4 +67,11 @@ void PopupPlacementAttached::apply() {
   m_window->setProperty("_q_waylandPopupAnchor", QVariant::fromValue(anchor));
   m_window->setProperty("_q_waylandPopupGravity", QVariant::fromValue(gravity));
   m_window->setProperty("_q_waylandPopupConstraintAdjustment", SLIDE_X | SLIDE_Y | FLIP_Y);
+
+  // Anchor to the trigger's real rect instead of letting QtWayland guess it
+  // from the (possibly shifted/wider) popup window geometry.
+  if (auto *trigger = parent()->property("parent").value<QQuickItem *>(); trigger && trigger->window()) {
+    const QRect rect = trigger->mapRectToScene(trigger->boundingRect()).toRect();
+    m_window->setProperty("_q_waylandPopupAnchorRect", rect);
+  }
 }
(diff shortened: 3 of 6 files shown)


src/server/src/qml/qml/SearchableDropdown.qml
```diff
@@ -95,8 +99,8 @@ Item {
     CompletionPopup {
         id: completionPopup
         parent: triggerButton
-        popupType: Popup.Window
-        x: compact ? triggerButton.width - width : 0
+        popupType: Popup.Item
```
aurelleb: this breaks on niri (triggers a drop shadow bug and removes blur as it's no longer a native popup

aurelleb: Image

aurelleb (changes requested):

src/server/src/qml/qml/SearchableDropdown.qml
pavle99: Thanks for checking, I'll keep the Popup.Window path on niri. The positioning fix now only uses Popup.Item on non-niri compositors, while niri keeps the native popup path to preserve blur/drop-shadow behavior. I tested the non-niri path locally on KDE; I can't run niri on my machine, but this should avoid the regression you pointed out because niri no longer takes the Popup.Item path. Still, please double check on your end.

aurelleb: no I don't think a per platform fix is the right solution, this highlights a deeper problem. This may create issues on other desktops, I just haven't tried.

aurelleb: no I don't think a per platform fix is the right solution, this highlights a deeper problem. This may create issues on other desktops, I just haven't tried.

aurelleb: no I don't think a per platform fix is the right solution, this highlights a deeper problem. This may create issues on other desktops, I just haven't tried.

pavle99: You're right. I tried removing the per-platform branch and keeping Popup.Window with generic overlay-space positioning, but Qt/Wayland still places the native popup outside the settings window. The only version that fixes placement uses Popup.Item, which regresses Niri blur/drop-shadow as you pointed out. I don't think this PR is the right fix anymore, unless you have a way to salvage it so this placement is proper on all environments.

pavle99:
Good news: I said earlier I couldn't salvage this, but I kept digging and found the actual root cause, so I've reworked the PR from scratch.

For popupType: `Popup.Window`, QtWayland builds the xdg_positioner from the popup's extended window type - only Qt's own `ComboBox/Menu/ToolTip` get below-the-control anchoring, while a plain Popup falls into the default top-right anchor on the trigger rect. The QML x/y are never part of the positioner (see `QWaylandXdgSurface::createPositioner` in qtbase). That's why every coordinate-based attempt failed, and why it looked fine on niri: tiled windows touch the screen edge, so `slide_x` happened to push the popup back over the trigger.

The fix keeps native popup windows (blur/drop-shadow untouched) and adds a small `PopupPlacement` attached type that opts a popup into the same compositor-side placement Qt uses for `ComboBox`, via the `_q_waylandPopup*` override properties QtWayland reads at popup-surface creation. No compositor detection - it's the standard xdg-positioner protocol, so it behaves identically on any compositor. One caveat worth flagging: those properties are a semi-private QtWayland mechanism (the source calls it the "properties fauxAPI"); it's been stable for years, but long-term this is arguably a Qt bug worth reporting upstream.

Also applied to `ViciToolTip` and `ShortcutRecorderField`, which had the same latent misplacement. Verified on KWin with WAYLAND_DEBUG (correct anchor/gravity sent, popups configured flush below/above their triggers).

Would appreciate a quick check on niri.

src/server/src/qml/qml/SearchableDropdown.qml
pavle99: @aurelleb Found the actual root cause and reworked the PR, see the new comment in the conversation.

dagimg-dot:
> Good news: I said earlier I couldn't salvage this, but I kept digging and found the actual root cause, so I've reworked the PR from scratch.
> 
> For popupType: `Popup.Window`, QtWayland builds the xdg_positioner from the popup's extended window type - only Qt's own `ComboBox/Menu/ToolTip` get below-the-control anchoring, while a plain Popup falls into the default top-right anchor on the trigger rect. The QML x/y are never part of the positioner (see `QWaylandXdgSurface::createPositioner` in qtbase). That's why every coordinate-based attempt failed, and why it looked fine on niri: tiled windows touch the screen edge, so `slide_x` happened to push the popup back over the trigger.
> 
> The fix keeps native popup windows (blur/drop-shadow untouched) and adds a small `PopupPlacement` attached type that opts a popup into the same compositor-side placement Qt uses for `ComboBox`, via the `_q_waylandPopup*` override properties QtWayland reads at popup-surface creation. No compositor detection - it's the standard xdg-positioner protocol, so it behaves identically on any compositor. One caveat worth flagging: those properties are a semi-private QtWayland mechanism (the source calls it the "properties fauxAPI"); it's been stable for years, but long-term this is arguably a Qt bug worth reporting upstream.
> 
> Also applied to `ViciToolTip` and `ShortcutRecorderField`, which had the same latent misplacement. Verified on KWin with WAYLAND_DEBUG (correct anchor/gravity sent, popups configured flush below/above their triggers).
> 
> Would appreciate a quick check on niri.

image


Tested on niri and its working !

pavle99:
@dagimg-dot Thanks for checking it!

aurelleb:
Looks like this is the best fix we have for this for now, some KDE apps do it as well.

PR #1490, opened 2026-06-07
feat: Add footer status command menu

pavle99:
- Make the footer status area clickable for active commands.
- Add a lean command menu for preferences, alias, favorites, and deeplink copying.
- Add extension-specific store, path, and uninstall actions when available.

## Accessibility

Keyboard focus order is unchanged. The footer status remains a pointer-only `MouseArea` and is not added to the tab focus chain.


arturict:
Menu/footer command addition: if keyboard focus order changed, mention accessibility checks in PR description.

pavle99:
Addressed in the PR description: the footer status remains outside the tab focus chain, so keyboard focus order is unchanged.

PR #1506, opened 2026-06-11
qalc backend: improve question formatting, remove duplicated loads and parsing

leiserfg:
The duplicated loads where also in wrong order, that was what was breaking the `today + 1day` thingy


aurelleb:
looks like this broke a few other things

image

leiserfg:
A few flags later

image

PR #1510, opened 2026-06-12
feat: file search v2

aurelleb:
new file search implementation. highlights:
- fuzzy-ish queries
- attempted typo correction for queries with no successful matches
- full paths are searchable, instead of only filename
- better, more exhaustive watching of important directories
- visual feedback in vicinae when the file indexer is doing important work
- category filter
- uses less resources, especially during first indexing


src/file-indexer/src/important-dir-watcher-linux.cpp
```diff
@@ -0,0 +1,276 @@
...
+        handleEvent(*ev);
+      }
+    }
+  }
+
+  void eventLoop() {
+    pollfd fds[2] = {{m_inFd, POLLIN, 0}, {m_wakeFd, POLLIN, 0}};
+
+    while (true) {
+      if (poll(fds, 2, -1) < 0 && errno != EINTR) break;
```
porridgewithraisins: should be

```cpp
if (poll(fds, 2, -1) < 0) {
  if (errno != EINTR) break;
  else continue;
}
//...
```

cannot check fds[1].revents and such if the poll failed.

PR #1515, opened 2026-06-14
feat: add drag-and-drop support to SearchFilesView

### LINKED ISSUE #199: [Feature Request] Implement Drag and Drop
I'm not a Raycast user, but I believe Raycast supports **drag and drop** functionality.  

This can be especially useful with extensions like **SVGL**, where you often need to import SVGs into a browser or file manager by dragging and dropping, rather than copy-pasting.

porridgewithraisins:
Closes #199 

Disclaimer: vibe-assisted

Expose filePath and fileUrl model roles via SectionListModel enum so file results in the dedicated search view can be dragged to other applications. Uses DragHandler + Drag.Automatic with text/uri-list MIME type for external file drag-and-drop.

### DIFF
--- a/src/server/src/qml/qml/ListItemDelegate.qml
+++ b/src/server/src/qml/qml/ListItemDelegate.qml
@@ -13,6 +13,32 @@ SelectableDelegate {
     required property bool itemIsActive
     property var itemAccessory: []
     property string itemAccessoryColor: ""
+    property string filePath: ""
+    property string fileUrl: ""
+
+    readonly property bool _isDraggable: root.filePath !== ""
+
+    Component.onCompleted: {
+        if (_isDraggable) {
+            console.log("[DRAG] ListItemDelegate created: filePath=" + root.filePath + " fileUrl=" + root.fileUrl)
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


src/server/src/qml/qml/ListItemDelegate.qml
```diff
@@ -13,6 +13,32 @@ SelectableDelegate {
...
     property var itemAccessory: []
     property string itemAccessoryColor: ""
+    property string filePath: ""
+    property string fileUrl: ""
+
+    readonly property bool _isDraggable: root.filePath !== ""
+
+    Component.onCompleted: {
+        if (_isDraggable) {
+            console.log("[DRAG] ListItemDelegate created: filePath=" + root.filePath + " fileUrl=" + root.fileUrl)
```
aurelleb: do not leave log or use `console.debug`

aurelleb (changes requested):
small nit, good otherwise

src/server/src/qml/qml/ListItemDelegate.qml
porridgewithraisins: Ok

Issue #1463, opened 2026-05-31
Icons do not load


KimpiegamesYT1:
Since today the icons no longer are loading. They alls how questionmarks. 

<img width="974" height="636" alt="Image" src="https://github.com/user-attachments/assets/3f927b5c-0047-4765-b5a2-a4cecea4ae08" />

Version: 

<img width="546" height="518" alt="Image" src="https://github.com/user-attachments/assets/28369521-3c63-4c64-832f-50716d60fceb" />

Fedora KDE 44

aurelleb:
any logs?

lunoxide:
Not the original person but I am having a very similar issue.

<img width="860" height="540" alt="Image" src="https://github.com/user-attachments/assets/a7790cd2-3c73-4faf-83d5-f73739fecf75" />

```
[V] 2026-05-31T20:03:41 info  - Activated clipboard server "data-control"
[V] 2026-05-31T20:03:41 info  - Killed existing extension manager instance
[V] 2026-05-31T20:03:41 info  - Started extension manager /run/user/1000/vicinae/extension-manager.js with /usr/bin/node
[V] 2026-05-31T20:03:41 warn  - Failed to parse news state: "file_open_failure"
[V] 2026-05-31T20:03:41 info  - Starting clipboard server "data-control"
f[V] 2026-05-31T20:03:41 info  - Clipboard server "data-control" started successfully.
[V] 2026-05-31T20:03:41 info  - Starting incremental scan for entrypoint /home/user
[V] 2026-05-31T20:03:41 info  - Started "Qalculate!" calculator backend
[V] 2026-05-31T20:03:41 info  - Refreshing Qalculate exchange rates...
[V] 2026-05-31T20:03:41 info  - Anonymous telemetry is enabled. Learn more: https://docs.vicinae.com/telemetry
[V] 2026-05-31T20:03:41 info  - Vicinae server successfully started. Call "vicinae toggle" to toggle the window
ire[V] 2026-05-31T20:03:42 info  - Done walking file tree at /home/user. Processed 25245 directories and 131349 files in 0 seconds.
[V] 2026-05-31T20:03:42 warn  - QObject: Cannot create children for a parent that is in a different thread.
(Parent is QGuiApplication(0x7ffef5ba5af0), parent's thread is QThread(0x562edb1f3360), current thread is QThreadPoolThread(0x562edbc169e0)
[...]

PR #1516, opened 2026-06-14
fix(icons): identify icon themes by directory id, not display name

phantomcortex:
> [!NOTE]
> This patch was written mainly by Claude Opus 4.8 
> Please let me know if there's any issue with that. 

I believe this may solve #1463 

This has been bugging me for a bit and it's an issue that seems to persist through vicince versions. Tested with [Fluent](https://github.com/vinceliuice/Fluent-icon-theme) icons, [papirus](https://github.com/PapirusDevelopmentTeam/papirus-icon-theme) icons, Breeze, Breeze Dark, and my personal favorite [kora](https://github.com/bikass/kora). All of the themes listed at the table below are manually confirmed to not work correctly. I also added something to ignore cursor themes in the theme menu. 

Let me know if I missed anything or if anything doesn't meet standards.

---

The patch was tested Below you can see the before and after:

vicinae 0.21.6:
Screenshot From 2026-06-13 12-35-00 (Edited)
vicinae 0.21.6 `fix/icon-theme-directory-id`:
Screenshot From 2026-06-13 12-35-37 (Edited)

---

> [!NOTE]
> AI summary below because I think it explains things better than me:

# fix(icons): identify icon themes by directory id, not display name

## Summary

Custom icon themes selected in settings never actually applied. The icon-theme
database identified themes by their human-readable **display name** (the `Name`
key from `index.theme`, e.g. `"Kora"`), stored that as the config value, and
handed it to `QIcon::setThemeName`. But Qt resolves themes by their **directory
basename** (e.g. `"kora"`), so the lookup silently failed and no icons rendered.

This PR makes the **directory id** the canonical identifier everywhere, keeps the
display name strictly for presentation, migrates legacy configs that stored the
old display name, and makes `auto` mode actually follow the system icon theme on
Linux.

### Why the id, not the name (or a case-insensitive match)

The directory basename and the `index.theme` `Name` diverge in more ways than just
casing, so neither a case-insensitive match nor a special-cased override would be
robust. A survey of installed themes on one machine:

| Directory (what Qt resolves) | `index.theme` Name | Nature of the mismatch |
|---|---|---|
| `kora` | `Kora` | case only |
| `breeze` | `Breeze` | case only |
| `breeze-dark` | `Breeze Dark` | case + hyphen→space |
| `Fluent-dark` | `Fluent dark` | hyphen→space |
| `Fluent-light` | `Fluent light` | hyphen→space |

A case-insensitive match would fix `kora`/`Kora` but still break on `Fluent dark`
vs `Fluent-dark` (space vs hyphen). The only value Qt can actually resolve is the
directory basename, so that is what we treat as canonical.

### `internal/icon-theme-db/icon-theme-db.hpp` — model: split id vs. name

- Added `IconThemeInfo::id` (directory basename) alongside the existing `name`
  (display name), with doc comments spelling out that `id` is what
  `QIcon::setThemeName` requires and `name` is presentation-only.
  *Why:* the whole bug stems from conflating these two values. Making them
  distinct fields forces every call site to be explicit about which one it wants.
- Renamed `hasTheme` → `hasThemeId` and re-documented `isSuitableTheme`.
  *Why:* the lookup semantics changed from "match by name" to "match by id";
  renaming prevents a caller from accidentally passing a display name.
- Added `resolveThemeId(value)`, `displayName(id)`, and `systemIconThemeId()`
  declarations (see `.cpp` rationale below). Added `#include <optional>` for the
  new optional return type.

### `internal/icon-theme-db/icon-theme-db.cpp` — id-based scan, migration, system detection

- `scan()` now records `info.id` from the directory basename, falls back to the
  id when `Name` is missing, and **deduplicates on id rather than name**.
  *Why:* the same theme can appear under several search paths (system + user);
  the directory id is the stable key. Deduping on display name could drop a
  legitimately distinct theme that happens to share a `Name`.
- `scan()` also **skips entries with no `Directories` key** in `[Icon Theme]`.
  *Why:* cursor themes (e.g. `capitaine-cursors`) live under the same icon search
  paths but, per the XDG icon theme spec, omit the required `Directories` key.
  Filtering on its absence keeps cursor themes (and other non-icon entries) out of
  the icon theme dropdown, where selecting one would render no icons. We test with
  `QSettings::contains` rather than reading the value, since `Directories` is a
  comma-separated list that `QSettings` would parse into a `QStringList`.
- `guessBestTheme()` now (1) consults `systemIconThemeId()` first, then (2) falls
  back to a well-known list matched **by id** (`Adwaita`, `breeze`, `Breeze`,
  `oxygen`), then any non-hidden theme, then `hicolor`.
  *Why:* the previous well-known list held display names that no longer match the
  id-based lookup; and following the actual system setting is far better than
  guessing.
- Added `resolveThemeId()` — returns a valid id as-is, otherwise maps a legacy
  display name back to its id, otherwise returns the input unchanged.
  *Why:* users who selected a theme before this fix have the old display name
  persisted in their config. This migrates them transparently with no data loss
  and no breakage if the value is already correct or the theme is uninstalled.
- Added `displayName()` — id → display name for the settings UI.
- Added `systemIconThemeId()` (Linux only) — reads the desktop's configured icon
  theme from KDE `kdeglobals` (`[Icons]/Theme`) on Plasma, then GNOME/GTK
  `gsettings`, then GTK `settings.ini` as a fallback. Every candidate is
  validated against installed themes before being returned.
  *Why:* `auto` should mean "match my desktop". Qt's own platform detection is
  unreliable under Wayland, so we read the source of truth directly. Validation
  ensures we never set a theme that isn't actually installed.

### `qml/general-settings-model.cpp` — dropdown stores id, displays name

- `iconThemeItems()` now builds dropdown entries as `makeDropdownItem(theme.id,
  theme.name)` — value = id, label = display name.
  *Why:* the value persisted to config must be the id so it round-trips to
  `QIcon::setThemeName` correctly, while the user still sees the friendly name.
- `currentIconTheme()` resolves the stored config value through `resolveThemeId`
  and renders its `displayName`.
  *Why:* keeps the selected entry correct even for legacy (display-name) configs,
  and shows the friendly label for the resolved id.

### `server.cpp` — apply themes by id, follow system in `auto`

- Both the config-reload handler and the `colorSchemeChanged` handler now, on
  Linux, run the concrete value through `resolveThemeId()` (handles legacy
  configs) and, for `auto`, call `guessBestTheme()` so the system theme is
  honoured. Non-Linux keeps the prior direct `setThemeName` path under `#else`.
  *Why:* this is where the id actually reaches Qt; without `resolveThemeId` here
  the migration would be incomplete, and routing `auto` through `guessBestTheme`
  is what makes "follow the system" work on light/dark switches and reloads.

## Testing

- Built in Arch linux container, Tested on GNOME 50 
- Selecting a custom theme (e.g. Kora) now renders its icons immediately.
- Configs created before this change (storing the display name) resolve correctly
  on next load with no manual intervention.
- `auto` follows the system icon theme and updates across light/dark switches.
- Cursor themes installed alongside icon themes no longer appear in the dropdown.

## Notes / out of scope

- The settings dropdown still has no explicit **"System default" (auto)** entry,
  so a user who picks a concrete theme can't re-select auto from the UI. This is
  pre-existing behaviour; happy to add it here if reviewers prefer.
- `systemIconThemeId()` calls `gsettings` synchronously with a 1s timeout on
  config/colour-scheme changes (not a hot path). Can be cached if it ever shows
  up as a stall.


arturict:
Theme/icon mapping change: sample extension directories for id/display name mismatch would aid validation.

PR #1529, opened 2026-06-18
feat: setWallpaper api

aurelleb:
a simple API to set the current wallpaper using whatever backend is available.

Needs some testing on:
- [ ] Gnome
- [x] KDE
- [x] macOS


dagimg-dot:
> a simple API to set the current wallpaper using whatever backend is available.
> 
> Needs some testing on:
> - [ ] Gnome
> - [ ] KDE
> 

Does this need a change from the extension? (For gnome)

Edited: no need, I think it works on normal gnome with gsettings

aurelleb:
@dagimg-dot yeah no it works with gsettings, unless the extension can do it better idk

dagimg-dot:
> @dagimg-dot yeah no it works with gsettings, unless the extension can do it better idk

Maybe per monitor wallpaper and other hidden apis

aurelleb:
well, if you want to implement this later down the line the  feel free to, it's just a backend change after all, won't impact the user facing API for extensions

PR #1545, opened 2026-06-23
build(nix): support macOS builds and allow system cmark-gfm / syntax-highlighting

natsukium:
I've been following this project for half a year now, and the recent pace of macOS support has been awesome to watch! I work across both Linux and macOS, so being able to use the same great launcher on every platform makes me really happy. Thank you!

## What this PR does

This PR extends the build definitions so that macOS users on Nix can try Vicinae easily.

In addition, the current `CMakeLists.txt` always fetched `cmark-gfm` and the KDE syntax-highlighting library via `FetchContent` on macOS. This PR lets macOS link against the system libraries the same way Linux already does, so the Nix derivation can provide them instead of vendoring.

If there's a cleaner way to handle any of this (especially the CMake option gating and the `.app` bundle assembly), I'd like to hear it.

## Follow-up

This PR focuses on packaging the build. In a follow-up PR I'm planning to add launchd support to the home-manager module.

Thanks again for developing such a nice piece of software!

### DIFF
--- a/nix/vicinae.nix
+++ b/nix/vicinae.nix
@@ -5,15 +5,36 @@
   kdePackages,
   lib,
   libqalculate,
+  llvmPackages_21,
   ninja,
   nodejs,
   npmHooks,
   pkg-config,
   qt6,
+  stdenv,
   gcc15Stdenv,
   wayland,
   glaze,
 }: let
+  inherit (stdenv.hostPlatform) isLinux isDarwin;
+
+  effectiveStdenv =
+    if isLinux
+    then gcc15Stdenv
+    else llvmPackages_21.libcxxStdenv;
+
+  # syntax-highlighting (and its extra-cmake-modules dep) is tagged Linux-only in
+  # nixpkgs but builds fine on macOS; lift the platform tag so we can link the
+  # system library instead of vendoring it. No-op on Linux.
+  addDarwinPlatform = drv:
+    drv.overrideAttrs (old: {
+      meta = old.meta // {platforms = old.meta.platforms ++ lib.platforms.darwin;};
+    });
+  kdeScope = kdePackages.overrideScope (_: prev: {
+    extra-cmake-modules = addDarwinPlatform prev.extra-cmake-modules;
+  });
+  syntax-highlighting = addDarwinPlatform kdeScope.syntax-highlighting;
+
   manifestRaw = builtins.readFile ../manifest.yaml;
   manifestGet = key: let
     m = builtins.match ".*${key}:[[:space:]]*\"([^\"]+)\".*" manifestRaw;
@@ -22,7 +43,7 @@
     then throw "Key ${key} not found in manifest.yaml"
     else builtins.elemAt m 0;
 in
-  gcc15Stdenv.mkDerivation (finalAttrs: {
+  effectiveStdenv.mkDerivation (finalAttrs: {
     pname = "vicinae";
     version = lib.removePrefix "v" (manifestGet "tag");
 
@@ -43,7 +64,10 @@ in
       "VICINAE_GIT_COMMIT_HASH" = manifestGet "short_rev";
       "VICINAE_PROVENANCE" = "nix";
       "INSTALL_NODE_MODULES" = "OFF";
+      "USE_SYSTEM_CMARK_GFM" = "ON";
       "USE_SYSTEM_GLAZE" = "ON";
+      "USE_SYSTEM_KF6" = "ON";
+      "USE_SYSTEM_QT_KEYCHAIN" = "ON";
       "CMAKE_INSTALL_PREFIX" = placeholder "out";
       "CMAKE_INSTALL_DATAROOTDIR" = "share";
       "CMAKE_INSTALL_BINDIR" = "bin";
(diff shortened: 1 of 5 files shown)


aurelleb:
thank you!
I may vendor cmark-gfm eventually as it's causing a few problems, and it hasn't been updated in years.

@cswimr @vMohammad24 if you guys can review this PR, it would be helpful, thx!

vMohammad24:
LGTM on the Linux side. 

everything looks fine on Darwin but if you have nix on your Mac you should test it.

cswimr (commented):
I don't have a Darwin machine to test on, so i can't give a ton of insight, but this all looks good to me from a Nix standpoint. If it hasn't already been tested, someone needs to make sure this builds on Darwin. I have tested evaluation on Linux, and it's fine. Maybe we should consider making this part of the repository's CI? I think it's probably pretty unlikely that future contributors to the Nix package will make sure the Darwin build passes, considering the overwhelming majority of them will be using Linux.

nix/vicinae.nix
```diff
@@ -85,21 +109,41 @@ in
...
+    ] ++ lib.optionals isLinux [
       "--set VICINAE_INPUT_SERVER_BIN /run/wrappers/bin/vicinae-input-server"
     ];
 
     meta = {
       description = "A focused launcher for your desktop — native, fast, extensible";
       homepage = "https://github.com/vicinaehq/vicinae";
       license = lib.licenses.gpl3Plus;
-      platforms = lib.platforms.linux;
+      platforms = lib.platforms.linux ++ lib.platforms.darwin;
```
cswimr: minor nit: this seems a little awkward, consider using a `with` instead? This is the stylistic standard in nixpkgs from what I've seen.
```suggestion
      platforms = with lib.platforms; linux ++ darwin
```

dnlzro:
Builds, runs, and integrates with Home Manager without issue for me. Some features are missing (most notably, clipboard history) but I don't think that's related to the Nix packaging.

aurelleb:
@dnlzro yeah the macOS release is basically in alpha state rn, there are many features that are not working. Thank you for testing on a mac.

nix/vicinae.nix
natsukium: Thanks for the suggestion. I usually try to avoid `with` expressions, but since the scope is limited and unambiguous here, I agree it makes sense. I'll update it.

natsukium:
Thank you everyone for the review and for testing this on macOS!

Regarding the CI, since macOS support is still in its early stages, we can probably wait until the Homebrew-based CI is up and running. It already works fine for my daily use. I'll be building this software almost every day anyway, so I'll catch it quickly if anything breaks.

As for the `cmark-gfm` vendoring, please feel free to move forward with that whenever you're ready. From what I can see, the Nix-side adjustments should be identical for both Linux and Darwin. I've been maintaining CMake-based software in nixpkgs for nearly 4-5 years now, so I should be able to handle any fixes.

Thanks again for all the help!

natsukium:
Ah, my bad, I just noticed there's already a PR for macOS CI (#1556).

If hosting additional runners isn't an issue, setting that up would be amazing. It would be convenient for end-users since they could then fetch pre-built binaries directly from Cachix!

cswimr:
One more thing to mention; as of NixOS (& nixpkgs) 26.05, [support for `x86_64-darwin` has been deprecated](https://nixos.org/manual/nixpkgs/stable/release-notes#sec-nixpkgs-release-26.05-highlights). When 26.11 comes out, nixpkgs will no longer support building packages for that platform from source, nor will there be cached binaries. This raises the question; do we want to support Vicinae on Intel Macs until 26.11, or just go ahead and preemptively drop support? It's also worth noting that Homebrew will also be [demoting all Intel Macs to its "tier 3" support standard](https://docs.brew.sh/Support-Tiers#future-macos-support) by the time 26.11 releases.

Edit: The NixOS release blog post mentions this, but it's also worth noting that nothing prevents someone from installing a supported Linux distribution on Intel Macs, instead of using MacOS on them. Apple has stated that MacOS 26 Tahoe will be the final version of MacOS to support Intel Macs.

Edit (Again): As of writing, the currently ongoing staging cycle will be the final time Hydra will be building packages for `x86_64-darwin` for unstable branches. I'm not completely certain but I believe backports to 26.05 will continue being built, however Vicinae currently pins to `nixos-unstable`. With that in mind, I think it'd be wise to just drop `x86_64-darwin` completely, at least for the Nix package.

natsukium:
Thanks for the detailed context. I agree we have no need to provide legacy `x86_64-darwin` support, especially since we are just introducing macOS support now.

From what I see, there's nothing specific we need to do on the `vicinae` side anyway. If needed, upstream tools like `nix-systems/default` will handle the removal, and we can just inherit that behavior.

Also, regarding `meta.platforms`, it already includes platforms like RISC-V and PowerPC via `lib.platforms.linux`, but that doesn't mean we have to actively support all of them. It just indicates that the package can be built there. In the same sense, adding `lib.platforms.darwin` should be fine.

cswimr:
Wasn't aware that `lib.platforms.linux` contained so many different architectures, yeah I suppose leaving `.darwin` there is fine. Honestly had forgotten about the `systems` thing, that's my bad. Looks like `nix-systems` is planning on removing `x86_64-darwin` in December, when 26.05 reaches end-of-support. We have two options, then:

- Leave it as is, and just update the flake lockfile in December, dropping support for `x86_64-darwin` then.
- Pin to `nix-systems/default/future-26.11` or `nix-systems/triplet` now, and go back to `nix-systems/default` in December.

Either way we will have an additional commit to be made in December, whether that's a flake lockfile update or a change to inputs. I think we should probably just leave it as is and let it silently be removed in December.

natsukium:
I agree with the first option. Let's leave it as is for now and just update flake.lock in December.

Regarding supported systems, users on minor platforms generally expect that things might not work out of the box, so we don't need to worry too much about it. In fact, if we restrict the architectures too tightly, it forces users to use overrides, which makes it less convenient. Unless a package handles platform-specific binaries and absolutely cannot run elsewhere, it's better to keep the platform constraints relaxed.

Just to be sure, I tested building for `x86_64-darwin` using Rosetta, and it compiled cleanly without any issues. There is really no reason to preemptively restrict it right now.

aurelleb:
Regarding the GitHub workflow we just got sponsored by [depot.dev](https://depot.dev/) so we don't really have to think about CI/CD expenses too much. You can add a macOS nix workflow or modify the existing one, whatever is the simplest.

natsukium:
Thanks for the suggestion! I've added a macOS runner to the cachix workflow. 

Also, a big thank you to depot.dev for providing these awesome resources!

aurelleb:
@natsukium looks like the nix macOS workflow fails to install nix.
As a side note, I changed the workflow to also run on PRs now so if you open a new one we will be able to monitor this.
Thank you for your work!

PR #1547, opened 2026-06-23
feat: Fixed image support in clipboard for MacOS

Signynt:
image


Signynt:
Seems to be closed with #1637

PR #1551, opened 2026-06-24
feat: Added commands for searching different file types

Signynt:
image


arturict:
Search command addition: any indexing migration required? Include note if old search tokens behavior changes.

PR #1554, opened 2026-06-24
fix: fixed suggested apps not showing up in actions on MacOS

Signynt:
image


src/server/src/services/app-service/macos/mac-app-database.mm
```diff
@@ -285,7 +291,11 @@ void collectEmbeddedApps(const fs::path &bundle, std::vector<fs::path> &out) {
     return url ? [ws URLForApplicationToOpenURL:url] : nil;
   }
   case TargetKind::Uti: {
-    UTType *ut = [UTType typeWithIdentifier:toNSString(t.normalized)];
```
aurelleb, line 289: why do we need this? My understanding is that since you changed the action panel to use a path identifier we are not going to hit this.

If I understand correctly you are attempting to transform a mime type into an UTI? I would prefer if we could have a separate target kind for mime and a single function to do the mime to UTI conversion, it would make more sense. 

As said earlier this doesn't seem to be needed to fix this so maybe discard this for now, and we will see whether we need `findCuratedOpeners` to handle direct mime types or not.

Signynt:
Good point. I reverted the change to `mac-app-database.mm`, so it's back to the original. The single-line fix in `file-list-item.hpp` is enough to resolve the MacOS issue.

PR #1555, opened 2026-06-24
fix: file paste not working from clipboard on MacOS

Signynt:
Fixed file paste from clipboard not working on MacOS.
Does not change behavior on Linux.


aurelleb:
this will get fixed when I have the proper macos clipboard server working #1439 
there won't be any trace of text/uri-list left in the platform agnostic clipboard service

PR #1569, opened 2026-06-28
ci: bump install-nix-action v25 -> v31

natsukium:
v25 ships the Nix 2.19.2 installer, which assigns `_nixbld1` to UID 301. On macOS Sequoia (15) and later that UID is already taken by a system account, so the installer fails with `eDSRecordAlreadyExists` on the depot-macos-26 runner. v31 bundles a fixed installer that uses a non-colliding UID range.

Refs: https://github.com/NixOS/nix/issues/10892

fixes https://github.com/vicinaehq/vicinae/actions/runs/28321040747/job/83902909328


aurelleb:
perfect, thx!

PR #1591, opened 2026-07-02
docs: fix extension docs link in README.md

embik:
Super small contribution, but I noticed that this link was going to a 404 page. Hope this is useful even though it's only a "typo fix" style PR.


aurelleb:
it's always helpful to have links that aren't broken!

PR #1607, opened 2026-07-05
build(linux): allow skipping file installation

embik:
I'm one of the Gentoo chads called out here: https://github.com/vicinaehq/vicinae/blob/9878d64f2d964752a200647b7956afbbe51b288c/CMakeLists.txt#L26

I am admittedly new to writing ebuilds (the packaging instructions for Gentoo), but noticed that the unconditional attempt to install files into `CMAKE_INSTALL_PREFIX` (`/usr` in practice, I suppose) was violating the build sandbox for my ebuild.

Instead of letting cmake install those files (themes, `.desktop` files, icons, systemd unit), I would like the package manager to do it from `extra/` directly.

This adds `INSTALL_SHARED_FILES`, `ON` by default (so my little snowflake distribution doesn't break everyone else's build/install process), which can be toggled `OFF` to delegate installation of files to an external process. Hope this is useful, if not please feel free to close.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,12 @@ option(
 	but they are still required build the project."
 	ON
 )
+option(
+    INSTALL_SHARED_FILES
+    "Create directories in CMAKE_INSTALL_PREFIX and install shared files (themes, .desktop files,
+    systemd unit, kernel module configuration)."
+    ON
+)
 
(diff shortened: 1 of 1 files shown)


arturict:
For linux build changes, a quick note on required environment variables in manifest generation would help review and troubleshooting.

PR #1609, opened 2026-07-05
build: split embedded resources based on build conditions (extensions vs core)

embik:
I noticed that `resources.qrc` is only added to the sources when `TYPESCRIPT_EXTENSIONS` is passed to `cmake`. But that includes the default configuration file, and my build kept throwing

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Failed to open default config
zsh: abort (core dumped)  vicinae server --open --no-extension-runtime
```

because the default config was simply not embedded in the binary.

In addition, I noticed that `@EXTRA_DIR@` in the template did not exist, and I assume `@EXTRA_PATH@` was the intended variable.

### DIFF
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -44,7 +44,8 @@ if (UNIX AND NOT APPLE)
 ")
 endif()
 
-configure_file(resources.qrc.in resources.qrc)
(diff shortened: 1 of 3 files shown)


aurelleb:
that makes a lot of sense, to be honest I just forgot about that flag and broke it at some point. thanks!

PR #1614, opened 2026-07-06
feat: macOS onboarding process

aurelleb:
add light onboarding process so that macOS users can give the required permissions

we keep it disabled on nix as we expect most of this to be handled in a declarative manner

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,7 @@ option(ENABLE_CLANG_TIDY "Run clang-tidy during compilation" OFF)
 option(INSTALL_MODULES_LOAD_CONFIG "Install config files to auto load required kernel modules" ON)
 option(AUTO_INSTALL_BROWSER_MANIFESTS "Install per-user browser native messaging host manifests at server startup" ON)
 option(AUTO_ENABLE_AUTOSTART "Register the app to start at login on first launch (currently macOS only)" ON)
+option(ENABLE_ONBOARDING "Show the onboarding window on first launch (macOS only)" ON)
 
 set(VICINAE_BUNDLE_ID "com.vicinaehq.Vicinae" CACHE STRING "macOS app bundle identifier")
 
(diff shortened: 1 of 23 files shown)


dnlzro:
> we keep it disabled on nix as we expect most of this to be handled in a declarative manner

Unfortunately, it isn't possible to give macOS apps permissions declaratively. Apps must present permission dialogs, or else the user has to manually add the app/binary to the relevant section of System Settings (which is kind of annoying, since you have to find the right path in the Nix store).

I think we should either:

1. Remove the `ENABLE_ONBOARDING` flag, so even Nix users have a chance to grant permissions.

    If we go this route, we should consider adding a flag that disables the "Launch at login" toggle, since that can and should be handled by Nix / Home Manager. We could also just re-use the existing `AUTO_ENABLE_AUTOSTART` for this purpose.

2. Keep `ENABLE_ONBOARDING=OFF`, but still prompt for Accessibility permissions on startup if they haven't yet been granted.

Thoughts?

aurelleb:
@dnlzro I think what we could do is just show the permission page of the onboarding if it's not granted. This is what I started doing since accessibility permission became mandatory, I think we should just lift the nix exclusion in that particular case. Other onboarding steps remain unnecessary

dnlzro:
@aurelleb My only concern is that for binaries outside of a .app bundle, permissions are granted based on the path of the binary. So if there's any change to the Vicinae binary (due to an update, change in configuration, etc.), the permissions would need to be granted again.

If I understand correctly, the onboarding will never run if `onboarding.json` is already present, even if the permissions have gone stale. So that would need to be changed.

aurelleb:
> If I understand correctly, the onboarding will never run if onboarding.json is already present, even if the permissions have gone stale. So that would need to be changed.

That is what I changed (as part of a later PR), now it is shown again (the permission pane) if you miss the Accessibility permission. The others are technically optional.

> permissions are granted based on the path of the binary. So if there's any change to the Vicinae binary (due to an update, change in configuration, etc.), the permissions would need to be granted again.

Isn't there a way to setup self signing with nix? That way you don't have this problem

dnlzro:
Ah okay gotcha, in that case I think the only thing that needs to change is to remove the Nix exclusion like you said.

> Isn't there a way to setup self signing with nix? That way you don't have this problem

I will look into this. I think the Home Manager module would need to change so it launches Vicinae from the .app bundle instead of the bare binary. But I'm already working on improving the module a little bit so I'll add that to the list.

PR #1615, opened 2026-07-06
feat: chrome native-messaging for nixos

yechielw:
This pr adds native-messaging support for the browser extension

fixes https://docs.vicinae.com/browser-extension#installation:~:text=Known%20exceptions%20are%3A-,NixOS%20(for%20now),-AppImage%20install%20that

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -54,6 +54,13 @@ in {
       '';
     };
 
+    enableChromeIntegration = lib.mkOption {
+      default = true;
+      description = ''
+        Whether to install the messaging host so that the chrome extension <https://chromewebstore.google.com/detail/vicinae-integration/kcmipingpfbohfjckomimmahknoddnke> works.
+      '';
+    };
+
     systemd = {
       enable = lib.mkEnableOption "vicinae systemd integration";
 
@@ -260,6 +267,22 @@ in {
           }
         ))
       ];
+
+      programs.google-chrome = lib.mkIf (cfg.enableChromeIntegration && cfg.package != null) {
+        enable = lib.mkDefault true;
+        package = lib.mkDefault null;
+        nativeMessagingHosts = [
+          (pkgs.writeTextDir "etc/chromium/native-messaging-hosts/com.vicinae.vicinae.json" (
+            builtins.toJSON {
+              name = "com.vicinae.vicinae";
+              description = "Vicinae Native Messaging Host";
+              path = "${cfg.package}/libexec/vicinae/vicinae-browser-link";
+              type = "stdio";
+              allowed_origins = ["chrome-extension://kcmipingpfbohfjckomimmahknoddnke/"];
+            }
+          ))
+        ];
+      };
 
       systemd.user.services.vicinae = lib.mkIf (cfg.systemd.enable) {
         Unit = {


aurelleb:
@cswimr @vMohammad24 feel free to review

nix/module.nix
```diff
@@ -261,6 +268,22 @@ in {
         ))
       ];
 
+      programs.google-chrome = lib.mkIf (cfg.enableChromeIntegration && cfg.package != null) {
+        enable = lib.mkDefault true;
+        package = lib.mkDefault null;
```
cswimr, line 273: This shouldn't have side-effects. It should only touch `nativeMessagingHosts`, like the Firefox integration does.

cswimr (changes requested):
Almost looks good, but please reformat this. If memory serves, we use alejandra.

nix/module.nix
yechielw: I reason for this additon is that `programs.google-chrome.nativeMessagingHosts` doesnt take affact unlsess `programs.chrome.enalbe` is set to true.
I didn't want it to actualy manage chrome so i figured that the best route is to enalbe it by defualt whilce disable the package value.

cswimr: Does `programs.google-chrome.nativeMessagingHosts` have side effects outside of the Google Chrome module? If not, it not taking effect when the module isn't enabled doesn't matter. Also, setting the package to `null` will mean that people who don't explicitly set a chrome package will get a broken module. I don't think that'd even pass typechecking, might be wrong there though.

cswimr: Oh, I just noticed that this chrome integration is enabled by default. In that case, we absolutely shouldn't be implicitly enabling the chrome module just by importing the vicinae home-manager module, that's a huge side effect that could be quite annoying to debug.

yechielw: programs in home-manager are a bit diffenrent then in nixos.
programs generaly set xdg config symlinks, enalbe home pakcages etc.
in this case:
```nix
programs.google-chrome =  {
        enable = lib.mkDefault true;
        package = lib.mkDefault null;
        nativeMessagingHosts = ...;
```
a user that hasn't explistly configured chrome would have the following imact:
the program would not add the google-chrome package to path but would set the symlink from `$out/etc/chromium/native-messaging-hosts/com.vicinae.vicinae.json` to `${config.xdg.configHome}/google-chrome/NativeMessagingHosts/com.vicinae.vicinae.json`

https://github.com/nix-community/home-manager/blob/63d02d1c19f1c2b47a5bc7e55c620b6af7b218a6/modules/programs/chromium.nix#L334

# having htat said...

packaging the native host messeging config via the home manager module is a bit weird
The right way is to pakcage them directly in the main vicinae pakcage as demestrated in the following branch
users would still need to import vicinae package via programs.google-chrome.nativeMessagingHosts = [vicinae]; and same for firefox
https://github.com/yechielw/vicinae/tree/chrome-integration-2
this is doen by adding the following to the derivation
```nix
lib.optionalString isLinux ''
        install -d "$out/lib/mozilla/native-messaging-hosts"
        cat > "$out/lib/mozilla/native-messaging-hosts/com.vicinae.vicinae.json" <<EOF
        {
          "name": "com.vicinae.vicinae",
          "description": "Vicinae Native Messaging Host",
          "path": "$out/libexec/vicinae/vicinae-browser-link",
          "type": "stdio",
          "allowed_extensions": ["firefox@vicinae.com"]
        }
        EOF

        install -d "$out/etc/chromium/native-messaging-hosts"
        cat > "$out/etc/chromium/native-messaging-hosts/com.vicinae.vicinae.json" <<EOF
        {
          "name": "com.vicinae.vicinae",
          "description": "Vicinae Native Messaging Host",
          "path": "$out/libexec/vicinae/vicinae-browser-link",
          "type": "stdio",
          "allowed_origins": ["chrome-extension://kcmipingpfbohfjckomimmahknoddnke/"]
        }
        EOF
      '';
```
in this case building vicinae itslef would generate 
result/etc/chromium/native-messaging-hosts/com.vicinae.vicinae.json
and result/lib/mozilla/native-messaging-hosts/com.vicinae.vicinae.json 
which can be imported automaticly as done here or explicitly by the user

yechielw:
I am going to close this PR in exchange for this https://github.com/vicinaehq/vicinae/pull/1625

PR #1623, opened 2026-07-07
fix(xdg): append terminal arg value to flags ending with '='

stvsu:
Per the [xdg-terminal-exec spec](https://github.com/Vladimir-csp/xdg-terminal-exec): "If argument expects a value and is defined as ending with `=`, value should be appended to the same argument without a white space."

`XdgAppDatabase::launchTerminalCommand` always emits the flag and its value as two separate argv entries. Terminals that declare the `=` form break — ghostty ships `X-TerminalArgAppId=--class=`, so since #1522 started passing an app id, launching any `Terminal=true` desktop entry produces `--class= <app-id>` and ghostty rejects the bare app id argument with a "Configuration Errors" dialog (`cli:3:nvim.desktop: invalid field`). Ghostty only accepts `--flag=value`, so appending per the spec is required there.

Repro: set ghostty as the default terminal (xdg-terminals.list), launch any `Terminal=true` app (e.g. btop) from vicinae. On v0.21.7+ the ghostty config-error dialog appears on every launch.

Built locally on Arch (ghostty 1.3.1, which reproduces the dialog reliably).

### DIFF
--- a/src/server/src/services/app-service/xdg/xdg-app-database.cpp
+++ b/src/server/src/services/app-service/xdg/xdg-app-database.cpp
@@ -438,9 +438,19 @@ bool XdgAppDatabase::launchTerminalCommand(const std::vector<QString> &cmdline,
   std::ranges::for_each(exec | std::views::drop(1), [&](auto &&arg) { argv << arg; });
   auto texec = getTermExec(*xdgApp);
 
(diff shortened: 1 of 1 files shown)


aurelleb:
that makes a lot of sense indeed, thank you!

PR #1625, opened 2026-07-07
feat(nix): build browser native host messaging into main derivation

yechielw:
This PR is in replacment of https://github.com/vicinaehq/vicinae/pull/1615
It does the following:
- builds the browser native host messaging config in the main derevation  
- adds support for chrome and chromium based browsers

### DIFF
--- a/nix/vicinae.nix
+++ b/nix/vicinae.nix
@@ -127,15 +127,41 @@ in
     # On macOS CMake installs only the CLI; assemble a thin .app bundle for the
     # GUI server. Both binaries go in the bundle so the CLI resolves the server
     # as a sibling (findServerBinary); $out/bin/vicinae is re-added in postFixup.
-    postInstall = lib.optionalString isDarwin ''
-      app=$out/Applications/Vicinae.app
-      install -Dm755 bin/vicinae-server "$app/Contents/MacOS/Vicinae"
-      install -Dm755 bin/vicinae "$app/Contents/MacOS/vicinae-cli"
-      install -Dm644 Info.plist "$app/Contents/Info.plist"
-      install -Dm644 ../extra/vicinae.icns "$app/Contents/Resources/vicinae.icns"
-      cp -r ../extra/themes "$app/Contents/Resources/themes"
-      rm -f "$out/bin/vicinae"
-    '';
+    postInstall =
+      lib.optionalString isDarwin ''
+        app=$out/Applications/Vicinae.app
+        install -Dm755 bin/vicinae-server "$app/Contents/MacOS/Vicinae"
+        install -Dm755 bin/vicinae "$app/Contents/MacOS/vicinae-cli"
+        install -Dm644 Info.plist "$app/Contents/Info.plist"
+        install -Dm644 ../extra/vicinae.icns "$app/Contents/Resources/vicinae.icns"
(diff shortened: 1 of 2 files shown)


yechielw:
@cswimr can you take a look?

nix/vicinae.nix
```diff
@@ -127,15 +127,41 @@ in
...
+        install -Dm755 bin/vicinae-server "$app/Contents/MacOS/Vicinae"
+        install -Dm755 bin/vicinae "$app/Contents/MacOS/vicinae-cli"
+        install -Dm644 Info.plist "$app/Contents/Info.plist"
+        install -Dm644 ../extra/vicinae.icns "$app/Contents/Resources/vicinae.icns"
+        cp -r ../extra/themes "$app/Contents/Resources/themes"
+        rm -f "$out/bin/vicinae"
+      ''
+      # Generate native host manifests for Firefox and Chromium, so the browser
+      # extention can communicate with the browser link
+      + lib.optionalString isLinux ''
```
cswimr: Why only on Linux?

cswimr (commented):
Thanks for working on this. Almost looks good, just one concern.

@vMohammad24, would you mind taking a look at this too?

nix/vicinae.nix
cswimr: Hm, I suppose the location for MacOS would be different, looks like browsers store native message hosts in different places on that platform. I'd still say that the build should include native messaging hosts on darwin, that way users have them defined already and just have to wire them up manually to wherever they're supposed to go.

nix/vicinae.nix
```diff
@@ -127,15 +127,41 @@ in
...
+        app=$out/Applications/Vicinae.app
+        install -Dm755 bin/vicinae-server "$app/Contents/MacOS/Vicinae"
+        install -Dm755 bin/vicinae "$app/Contents/MacOS/vicinae-cli"
+        install -Dm644 Info.plist "$app/Contents/Info.plist"
+        install -Dm644 ../extra/vicinae.icns "$app/Contents/Resources/vicinae.icns"
+        cp -r ../extra/themes "$app/Contents/Resources/themes"
+        rm -f "$out/bin/vicinae"
+      ''
+      # Generate native host manifests for Firefox and Chromium, so the browser
+      # extention can communicate with the browser link
```
vMohammad24: typo (extention -> extension)

nix/vicinae.nix
aurelleb: indeed. you can refer to this https://github.com/vicinaehq/vicinae/blob/e7e34695371737b371d6488c7f680b436d82093e/src/server/src/services/browser-extension/native-host-installer.cpp#L54

vMohammad24:
macos is possible to do in a similar way (if u change paths), otherwise LGTM.

yechielw:
Looking at the hm module’s source code, i see that it handles the path for
macos as well so i guess we can remove the linux only restriction

arturict:
Great command plumbing. Can you include a compatibility test for old and new command forms around cwd inheritance?

aurelleb:
is this ready?

yechielw:
works fine for me

cswimr (approved):
This looks good to me.

aurelleb:
needs formatting

yechielw:
> needs formatting

Done

PR #1631, opened 2026-07-08
Fix image scale in extension grid

EricKuck:
Grid images using the stretch or contain fit currently underfill their cells by what appeared to be about 50% on my macbook. This was especially obvious using the gif search extension for raycast, where it was difficult to even make out what the gifs it brought up were supposed to be. This PR makes all fit modes scale to the cell as documented.

I've collapsed the delegate's two ViciImage branches into a single `anchors.fill` image whose `fillMode` is driven by the fit mode:

- A fit of Contain gets a fill of PreserveAspectFit (scale to fit, letterboxed, never cropped, matches raycast)
- A fit of Fill gets a fill of PreserveAspectCrop (scale to fill, cropped)
- A fit of Stretch gets a fill of Stretch

From what I can tell, svg, emoji, etc, icon rendering is unchanged with this PR, as they were already decoded at the cell resolution and filled properly. Lower resolution rasters like gifs now scale up to fill instead of underfilling, which matches the documented raycast-like behavior.

Note that this also fixes Stretch, which was previously capped at native size and so didn't actually stretch small images.

Here's what the gif search extension looks like after this change:
image

### DIFF
--- a/src/server/src/qml/qml/ExtensionGridView.qml
+++ b/src/server/src/qml/qml/ExtensionGridView.qml
@@ -16,8 +16,6 @@ GenericGridView {
             readonly property var model: parent ? parent.cmdModel : null
             readonly property int sec: parent ? parent.cellSection : 0
             readonly property int itm: parent ? parent.cellItem : 0
-            readonly property bool _isFill: cellRoot.model ? cellRoot.model.fit === 1 : false
-
             readonly property string _cellColor: {
                 var _rev = cellRoot.model ? cellRoot.model.dataRevision : 0;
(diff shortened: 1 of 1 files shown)


aurelleb:
did you try the latest vicinae commit before this? I should have fixed at least part of this problem yesterday already.

EricKuck:
Oh nice, y'all got a macOS release out! Just tried it out and you're right, it does improve things:

image

However it still doesn't seem to scale as expected without my changes. Here's a screenshot of the same plugin in raycast:

image

aurelleb:
yeah you are right, going to make sure it doesn't break anything else but your fix seems to make a lot of sense at first glance

aurelleb:
yep, much better indeed, including on Linux. Thanks!
image

PR #1646, opened 2026-07-10
Fix copying multiple files in macOS (and probably X11 too?)

EricKuck:
On macOS and X11, if my code tracing is correct, copying multiple files to the clipboard manager of vicinae, then trying to paste it doesn't work at all. I see the `file://` urls displayed in the clipboard manager, but nothing pastes into Finder. It's because uri lists are supposed to be CRLF-separated rather than semicolon-separated. [source](https://www.rfc-editor.org/info/rfc2483/#section-5)


aurelleb:
The only problem with that logic is that `text/uri-list` could theoretically contain URIs that are not `file://` but I don't think that's something we would see very often.
And yes it's separated by CRLF and not `;`  good catch

PR #1649, opened 2026-07-11
nix: add module options for the NixOS module

cswimr:
This PR moves the pre-existing NixOS module to its own file, and adds `.enable` & `.package` options for it. The input server option defaults to being enabled, so this isn't a breaking change. It just allows people to disable the input server (and therefore disable needing to have VIcinae in their store, if they're not using the home-manager module) without having to remove the NixOS module from their imports.

Also renamed `module.nix` -> `home-manager-module.nix` for clarity, now that `nix/` has two modules.


nix/nixos-module.nix
```diff
@@ -0,0 +1,28 @@
...
+  config,
+  ...
+}: let
+  cfg = config.services.vicinae-input-server;
+in {
+  options.services.vicinae-input-server = {
+    enable = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = "Enable Vicinae's input server. Required for some functionality such as clipboard / emojji pasting and snippets.";
```
vMohammad24: typo (emoji)

cswimr:
@vMohammad24 if you wouldn't mind reviewing this that'd be great

nix/nixos-module.nix
```diff
@@ -0,0 +1,28 @@
+self: {
+  pkgs,
+  lib,
+  config,
+  ...
+}: let
+  cfg = config.services.vicinae-input-server;
+in {
+  options.services.vicinae-input-server = {
```
vMohammad24: in my opinion having this under programs.vicinae.input-server would be way better

nix/nixos-module.nix
```diff
@@ -0,0 +1,28 @@
...
+    enable = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = "Enable Vicinae's input server. Required for some functionality such as clipboard / emojji pasting and snippets.";
+    };
+    package = lib.mkPackageOption self.packages.${pkgs.stdenv.hostPlatform.system} "default" {
+      extraDescription = "The Vicinae package to wrap.";
```
vMohammad24, line 16: when would this be used? I think it should just be .enable and have the same functionality as the old one.

cswimr: I don't see why we shouldn't expose .package. We expose it for the home-manager module, this is no different.

vMohammad24: true i guess

nix/nixos-module.nix
```diff
@@ -0,0 +1,28 @@
+self: {
+  pkgs,
+  lib,
+  config,
+  ...
+}: let
+  cfg = config.programs.vicinae-input-server;
```
vMohammad24: could you rename this to `programs.vicinae.input-server`, this is the idiomatic way to do it, it's better for users using repl, LSP and generally easier to find than the alternative (other projects do this as well (e.x: `programs.steam.gamescopeSession`))

cswimr: Well, `programs.steam` is its own module. `programs.vicinae` doesn't exist on the NixOS side of things. This is controlling the vicinae input server, not vicinae itself. I think keeping it under its own namespace makes more sense. Willing to change it though.

cswimr: I suppose putting it under `programs.vicinae.input-server` makes sense if we have to add more options to the nixos module in the future.

vMohammad24: > I suppose putting it under `programs.vicinae.input-server` makes sense if we have to add more options to the nixos module in the future.

yeah exactly, it doesnt really matter at the end both work the same its just more of a qoL thing, up to you. LGTM otherwise.

skifli:
I recently updated my flake.lock and the input server stopped working - after searching through the Home Manager nixOS options for Vicinae I didn't find anything, and it was only when I went through the Vicinae GitHub repository here that I realised that a new option for the input server had been added. However, I would like to ask whether part of the design is intended behaviour. 

```diff
 {
   inputs,
   userVars,
   ...
 }: {
+  imports = [
+    inputs.vicinae.nixosModules.default
+  ];
+
+  programs.vicinae.input-server.enable = true; # Needed for snippets
+
   home-manager = {
     sharedModules = [inputs.vicinae.homeManagerModules.default];

     users.${userVars.username} = {
      programs.vicinae = {
        enable = true;
      };
     };
   };
 }
```

Above, you can see a simplified example of how I configure Vicinae in my [nixfiles](https://github.com/skifli/nixos/blob/0b1034ecf50c4bebd5ebbad194515dea76275047/users/programs/launcher/vicinae.nix) and the Git diff after this update. My question is would it not be better to also add the input server option to the home manager configuration? Because now, unless I am misunderstanding something (which I may be as my nix knowledge is probably not as extensive as you guys), unlike before when the input server was automatically enabled for the home manager module, now it is not. And that means that it has to be enabled, but since it only exists as a root program, the extra nixosModules input has to be separately imported, and the input server has to be manually enabled on its own, completely separate from the home manager configuration which contains everything else.

Again, I might be completely misunderstanding the correct way to configure Vicinae - in which case do please correct me :heart: .

vMohammad24:
it's true by default, you only need to import the nixos module, it's the same exact behavior as before this PR.

skifli:
> it's true by default, you only need to import the nixos module, it's the same exact behavior as before this PR.

Oh yes of course, but still, before the PR the nixos module didn't have to be imported - now it does, which in my opinion is a bit confusing / cumbersome as one would be importing the nixos module to enable a function that doesn't exist in the home manager module, when all of the other configuration is done in the home manager module. But still, if it's intended behaviour it doesn't matter, just might be good to add to the docs. Thanks anyway.

vMohammad24:
it is in the docs, the nixos module is required so we can create a wrapper with permission to read and write user input, it's always been needed.  the only difference this PR introduces is that we now have an option to disable the input server while keeping the nixos module.

PR #1651, opened 2026-07-12
fix: accept object form of List/Grid `filtering` prop

### LINKED ISSUE #1581: List `filtering={{ keepSectionOrder }}` object form crashes extension worker (parseRenderPayload: expected_true_or_false)
**System information**

- Version: v0.22.0 (5d5a50db0)
- Provenance: nix
- OS: NixOS 26.11 (Zokor) - 26.11
- QT Platform: wayland
- DE: Hyprland/Wayland

**Describe the bug**

`parseRenderPayload` rejects the object form of the `List` `filtering` prop. The Raycast API types `filtering` as `boolean | { keepSectionOrder: boolean }`, but the parser only accepts a boolean — when a render payload contains `"filtering":{"keepSectionOrder":true}` it aborts with `expected_true_or_false`, the view never renders, and the extension worker exits with code 1.

This breaks several stock **Linear** extension (`store.raycast.linear`) commands that use the object form: **My Issues**, **Search Projects**, **Created Issues**, **Active Cycle**. **Search Issues** works because it passes `filtering={false}`.

**To Reproduce**

Minimal (no Linear needed) — a command rendering:

```tsx
import { List } from "@vicinae/api";

export default function Command() {
  return (
    <List filtering={{ keepSectionOrder: true }}>
      <List.Item title="test" />
    </List>
  );
}
```

The worker crashes on render. Or, via the store: install the Linear extension, log in, run **My Issues** or **Search Projects** → blank/closes, worker exits.

**Expected behavior**

The object form `filtering={{ keepSectionOrder: true }}` renders the same as `filtering={true}` (honoring or ignoring `keepSectionOrder`), instead of crashing the worker.

**Screenshots**

N/A — failure is in the render pipeline, no UI shown.

**Additional context**

Server logs on running **Search Projects**:

```
[E] Loaded extension store.raycast.linear:search-projects
[E] Started extension linear:search-projects
[V] parseRenderPayload: failed: "1:78: expected_true_or_false
    ...sLoading":true,"filtering":{"keepSectionOrder":true},"pagination...
                                      ^"
[E] Worker store.raycast.linear:search-projects exited with code 1
```

Upstream Raycast source uses the object form directly (e.g. `extensions/linear/src/assigned-issues.tsx`, `components/ProjectList.tsx`, `created-issues.tsx`, `active-cycle.tsx`), so it's valid extension code — the incompatibility is in Vicinae's parser. Distinct from #729 (that failure is `Action.PickDate` / React #130).

*(Bug report drafted with AI assistance — Claude Opus 4.8.)*

noamsto:
Fixes #1581.

`parseRenderPayload` only accepted a boolean for the `List`/`Grid` `filtering` prop, but Raycast types it as `boolean | { keepSectionOrder: boolean }`. The object form aborted with `expected_true_or_false`, so the view never rendered and the extension worker exited — breaking stock Linear commands that use it (My Issues, Search Projects, Created Issues, Active Cycle).

This types the wire `filtering` as `std::variant<bool, FilteringOptionsWire>` — mirroring the existing `ActionPannelSubmenu` filtering handling — and collapses it to the enabled bool. The object form enables filtering; `keepSectionOrder` is not honored yet (matches the boolean behavior).

Tested locally: `linux-debug` build passes, and `parseRenderPayload` now parses both `filtering: true` and `filtering: { keepSectionOrder: true }` for List and Grid into a real model instead of returning empty.

_Made with assistance from Claude (Opus 4.8)._

### DIFF
--- a/src/server/src/extend/model-deser.cpp
+++ b/src/server/src/extend/model-deser.cpp
@@ -320,9 +320,22 @@ template <> struct glz::meta<ListWireChild> {
       std::array{"list-item", "list-section", "action-panel", "empty-view", "dropdown"};
 };
 
+// Raycast types `filtering` as `boolean | { keepSectionOrder: boolean }`. The
+// object form enables filtering; keepSectionOrder is not honored yet.
+struct FilteringOptionsWire {
+  bool keepSectionOrder = false;
+};
+
+static bool resolveFiltering(const std::optional<std::variant<bool, FilteringOptionsWire>> &filtering,
+                             bool fallback) {
+  if (!filtering) return fallback;
+  if (auto *enabled = std::get_if<bool>(&*filtering)) return *enabled;
+  return true;
+}
+
 struct ListModelWire {
   bool isLoading = false;
-  std::optional<bool> filtering;
+  std::optional<std::variant<bool, FilteringOptionsWire>> filtering;
   bool throttle = false;
   bool isShowingDetail = false;
   std::string navigationTitle;
@@ -365,7 +378,7 @@ template <> struct glz::meta<GridWireChild> {
 
 struct GridModelWire {
   bool isLoading = false;
-  std::optional<bool> filtering;
+  std::optional<std::variant<bool, FilteringOptionsWire>> filtering;
   bool throttle = false;
   double aspectRatio = 1.0;
   std::optional<int> columns;
@@ -788,7 +801,7 @@ static ListModel toListModel(ListModelWire &w) {
   ListModel m;
   m.dirty = true;
   m.isLoading = w.isLoading;
-  m.filtering = w.filtering.value_or(!w.onSearchTextChange.has_value());
+  m.filtering = resolveFiltering(w.filtering, !w.onSearchTextChange.has_value());
   m.throttle = w.throttle;
   m.isShowingDetail = w.isShowingDetail;
   m.navigationTitle = std::move(w.navigationTitle);
@@ -854,7 +867,7 @@ static GridModel toGridModel(GridModelWire &w) {
   GridModel m;
   m.dirty = true;
   m.isLoading = w.isLoading;
-  m.filtering = w.filtering.value_or(!w.onSearchTextChange.has_value());
+  m.filtering = resolveFiltering(w.filtering, !w.onSearchTextChange.has_value());
   m.throttle = w.throttle;
   m.aspectRatio = w.aspectRatio;
   m.columns = w.columns;


arturict:
Could you add a tiny reproduction for object-form filtering behavior to reduce ambiguity in expected output?

noamsto:
@arturict Here is the expected `filtering` behavior for every input shape. `parseRenderPayload` resolves the wire `filtering` prop against the `onSearchTextChange` fallback via `resolveFiltering`:

| Payload `filtering`                     | `onSearchTextChange` | Resolved `filtering` |
|-----------------------------------------|----------------------|----------------------|
| `{"keepSectionOrder": true}` (object)   | any                  | `true`               |
| `{"keepSectionOrder": false}` (object)  | any                  | `true`               |
| `true`                                  | any                  | `true`               |
| `false`                                 | any                  | `false`              |
| absent                                  | absent               | `true`               |
| absent                                  | present              | `false`              |

The object form always enables filtering (matching Raycast, where the object is used to tweak filtering, not disable it); `keepSectionOrder` is parsed but not honored yet. The boolean form passes through unchanged, and the absent case keeps the prior default (filter locally unless the extension drives search via `onSearchTextChange`).

Before the fix, the object form hit `expected_true_or_false` and aborted the whole render, so the last two columns were the only reachable states.

aurelleb:
I think this guy is just spamming slop everywhere, please don't bother @noamsto

arturict:
That's fair feedback. I'm trying to move from wanting to contribute to OSS to actually doing useful work. I do use AI to assist with analysis and drafting, but I review and approve everything before posting. In this case, my comment was too generic and didn't add enough value, and that's on me.

I'll pause contributing here for now. If the maintainers are open to it, I'm happy to continue only when I can reproduce something and add concrete evidence. If they'd prefer that I move elsewhere, that's completely fine too — no hard feelings.

aurelleb:
one more llm message and you are getting permanently banned from contributing here I'm not kidding @arturict 
You clearly don't understand what is happening here and you are not helpful at all
image

arturict:
> one more llm message and you are getting permanently banned from contributing here I'm not kidding @arturict 
> You clearly don't understand what is happening here and you are not helpful at all
> 

Sorry man, English isn't my first language. I think its better that i stop. Good success with y'all's work.

aurelleb:
English is not my first language either, not an excuse. And you won't see me spam pull requests and issues with meaningless  messages like you did.

PR #1652, opened 2026-07-12
feat: adds the cwd to the extension launch context when the extension is launched from the cli

FredrikMWold:
remake of #1523

### DIFF
--- a/figura/manager-extension.fig
+++ b/figura/manager-extension.fig
@@ -33,6 +33,7 @@ struct LaunchEventData {
   support_path: string;
   preferenceValues: any;
   argumentValues: any;
+  launch_context: any;
   is_raycast: bool;
   owner_or_author_name: string;
   launch_type: LaunchType;
--- a/figura/manager.fig
+++ b/figura/manager.fig
@@ -38,6 +38,7 @@ struct LoadOptions {
   owner_or_author_name: string;
   arguments: any;
   preferences: any;
+  launch_context: any;
   launch_type: LaunchType;
   capabilities: Capabilities;
 };
--- a/src/cli/src/cli.cpp
+++ b/src/cli/src/cli.cpp
@@ -8,11 +8,32 @@
 #include "generated/version.h"
 #include "theme.hpp"
 #include "ipc-client.hpp"
+#include <filesystem>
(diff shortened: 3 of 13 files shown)


aurelleb:
Something I don't like with the current deeplink approach is that a deeplink is a generic concept that is not only used for launching commands. As a result the CLI interface for it feels a bit clumsy and undocumented (we pass things as url parameters).

I wonder if, for advanced CLI use like you do, it wouldn't be preferable to have a dedicated `vicinae launch` verb that can take domain specific flags (for example to override cwd manually if needed) so that we can specifically address launching commands as its own specific thing.

Deeplinks are still useful because they can be used as URIs but for advanced scripting needs I feel like they are starting to become a bit annoying.

What do you think about that?

FredrikMWold:
I think it would be good improvement. Allowing an args array would nice too.  That would make it easier to pipe stuff into it. Like git does to your editor command.

aurelleb:
can you elaborate on what you mean? can you give an example usage?

FredrikMWold:
The example would be to allow both 
`vicinae launch <extension>  <arg1> <arg2>`
`vicinae launch <extension>  --namedarg <value>` 

An example use case is for the "use vicinae as your git editor" extension i have.
When you call git commit, git uses your defined editor and calls it with a file path. 

With support for an arg array you would be able to set your editor to this and i should just work:
`vicinae launch git-editor`. It would then be called like this `vicinae launch git-editor <path-to-temp-commit-file>`

aurelleb:
I see that would be pretty cool indeed. However I'm not sure about how we would implement the positional argument thing, because arguments are identified by keys

FredrikMWold:
Would there be an issue with a simple solution like just assigning something like $1, $2, $3 ... as keys to unnamed arguments?

aurelleb:
Actually I just realized we specify arguments as an array in the manifest. So we can just assume positional arguments are given in order. Since they are also shown in order in the launcher UI I guess it makes sense. The CLI would return an error if one required argument is missing.

FredrikMWold:
That sounds good

arturict:
Would be helpful to include an example output for newly added launch context metadata when extension launched from CLI.

PR #1661, opened 2026-07-13
Adds new cmd verb to cli and attaches cwd to extension launch from it

### LINKED ISSUE #1657: "command" verbs in CLI

Currently the only way to start a vicinae command from the CLI is to use the deeplink which is a generic concept that also applies to standard controls / navigation. This makes it difficult to add domain specific parameters to it, as they have to be passed as URL parameters which are not easily discoverable

- `vicinae cmd ls` to list all loaded commands
- `vicinae cmd launch <entrypoint> [...args]` to launch a given command

`vicinae cmd launch` can take command arguments as follows:

Let's assume the command takes two strings as input, `foo` and `bar`:

```
# positional arguments after the entrypoint will be matched to every argument in order
vicinae cmd launch my-extension.foobar MyFoo MyBar
```

`vicinae cmd launch` may also accept arguments such as `--cwd` to forward a different current working directory to the command. The command should have a way to read what `cwd` the launch was initiated from. See #1652 

Deeplink usage, but too bare bones for advanced scripting needs.

FredrikMWold:
This PR adds a new `cmd` cli verb that lets the user launch a command from the cli easier. It also attaches the cwd to the a new launchContext for extensions.

image
image
image

Fixes #1657

### DIFF
--- a/figura/ipc.fig
+++ b/figura/ipc.fig
@@ -40,8 +40,12 @@ struct LaunchCommandResponse {
   error?: string;
 };
 
+struct CommandInfo {
+  id: string;
+};
+
 struct ListCommandsResponse {
-  commands: string[];
+  commands: CommandInfo[];
 };
 
 struct DMenuRequest {
--- a/figura/ipc.fig
+++ b/figura/ipc.fig
@@ -43,6 +43,7 @@ struct LaunchCommandResponse {
 
 struct CommandInfo {
   id: string;
+  name: string;
 };
 
 struct ListCommandsResponse {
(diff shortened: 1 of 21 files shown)


arturict:
Nice user-facing change. A quick screenshot or log excerpt around extension command output would help verify payload shape.

arturict:
Looks good. Can you validate CLI help text/docs mention the new --format index option and the output compatibility expectations (text vs index).

aurelleb:
@FredrikMWold I added the ability to list/launch all kinds of entrypoints including apps, scripts, shortcuts etc since I don't really see a valid reason to exclude them. They can all have arguments, and the launch deeplink supports them as well

FredrikMWold:
Why not, even better 😁

PR #1664, opened 2026-07-14
feat(settings): expose automatic window material

adamanz:
## Summary

- add **Automatic** to the window material picker
- show the configured automatic state instead of its resolved material
- clarify that automatic prefers Liquid Glass and falls back to blur

## Why

Selecting an explicit window material currently removes the only UI path back to the capability-aware `auto` behavior. This restores that option without changing existing defaults.

## Checks

- `cmake --preset macos-debug`
- `cmake --build --preset macos-debug`
- `make check-format`
- `scriptcommand-tests` (256 assertions)
- `vicinae-crypto-tests` (36 assertions)
- `vicinae-fuzzy-tests` (53 assertions)
- `vicinae-glyph-tests` (1,980 assertions)
- `xdgpp-tests` (257 assertions)

Prepared with Codex; locally reviewed and tested.

### DIFF
--- a/extra/config.jsonc
+++ b/extra/config.jsonc
@@ -108,7 +108,7 @@
 		},
 
 		// "none", "blur", "liquid_glass", or "auto" (liquid_glass when available, else blur).
-		// Needs window opacity < 1 to be visible.
+		// When opacity is unset, liquid glass uses a translucent default automatically.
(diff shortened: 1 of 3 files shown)


aurelleb:
is this ready, or still a draft?

PR #1676, opened 2026-07-16
fix(macos): prevent clipboard copy HUD stealing focus

Tinkerbells:
On macOS, when copying an item from Clipboard History, Vicinae shows a HUD saying "Selection copied to clipboard".

   Accessibility Inspector reports this HUD as a system dialog / AXSystemDialog with Keyboard Focused = true. This causes the currently focused app/window to lose focus
 temporarily. The issue is especially visible when using yabai/skhd: after copying from Vicinae Clipboard History, focus returns with a delay or appears blocked by the HUD.

   Expected behavior:
   The HUD should not accept keyboard focus, or Clipboard History copy should avoid showing a focusable HUD on macOS.

   Environment:
   - macOS
   - yabai + skhd
   - Vicinae v0.23.1

   Workaround tested:
   Building Vicinae with the macOS HUD window using Qt.WindowDoesNotAcceptFocus and suppressing the Clipboard History copy HUD on macOS resolves the focus issue.

### DIFF
--- a/src/server/src/qml/macos-chrome-attached.mm
+++ b/src/server/src/qml/macos-chrome-attached.mm
@@ -471,6 +471,8 @@ void MacOSPanelAttached::apply() {
   NSWindow *nswin = nsview.window;
   if (!nswin) return;
 
+  const bool acceptsFocus = !(m_window->flags() & Qt::WindowDoesNotAcceptFocus);
+
   if (!m_snapshot.valid) {
     m_snapshot.valid = true;
     m_snapshot.styleMask = (unsigned long)nswin.styleMask;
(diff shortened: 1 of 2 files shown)


src/server/src/extensions/clipboard/history/clipboard-history-actions.hpp
```diff
@@ -24,7 +25,11 @@ class CopyClipboardSelection : public AbstractAction {
     auto clipman = ctx->services->clipman();
     auto toast = ctx->services->toastService();
     if (clipman->copySelectionRecord(m_id, {.concealed = true})) {
+#ifdef Q_OS_MACOS
```
aurelleb: why do we need to do this if we prevent the HUD from grabbing focus? I would still want the HUD to show on macOS.

Tinkerbells: You're right. I removed the macOS-specific suppression of the clipboard HUD.
The updated patch keeps the HUD visible and only prevents it from accepting keyboard focus by adding Qt.WindowDoesNotAcceptFocus to the macOS HUD window and making the native NSPanel setup respect that flag.

PR #1682, opened 2026-07-16
Add launchd (macOS) support to home-manager module

dnlzro:
Add new `launchd` options in the Home Manager module to enable automatic startup on Nix Darwin systems, as well the ability to pass custom environment variables to the Vicinae process (if desired).

i.e., `launchd` is `systemd` for macOS. It is configured very similarly through Home Manager, so I extracted some helper functions to avoid code duplication in the module. 

With this addition, the module now works as I'd expect it to work. Vicinae launches at startup, watches the clipboard, is configured according to `programs.vicinae.settings`, etc.

**Note:** This PR builds on the work already done by @natsukium to support Nix Darwin (thank you! ❤️). The codebase has changed a decent amount since they last worked on their Home Manager module, and I just wanted to simplify things a bit and push this over the finish line.

### DIFF
--- a/nix/home-manager-module.nix
+++ b/nix/home-manager-module.nix
@@ -17,6 +17,18 @@ self: {
 
   jsonFormat = pkgs.formats.json {};
   tomlFormat = pkgs.formats.toml {};
+
+  envVarType = with lib.types; attrsOf (oneOf [str int float bool]);
+
+  envValueToString = val:
+    if lib.isBool val
+    then
+      (
+        if val
+        then "1"
+        else "0"
+      )
+    else toString val;
 in {
   disabledModules = ["programs/vicinae"];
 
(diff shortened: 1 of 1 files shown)


aurelleb:
@cswimr @vMohammad24 feel free to review

cswimr (approved):
This all looks good to me.

vMohammad24:
LGTM, limiting `launchd.agents.vicinae` to an if statement to put it behind darwin would be beneficial (although if it is enabled it would be the user's fault, not ours).

dnlzro:
Internally, Home Manager has its own assertion to make sure the host `isDarwin` when you specify `launchd` options. But I'm happy to guard it with `mkIf` anyway, if you'd prefer --- in that case, `programs.vicinae.systemd` should probably also get the same treatment (with `isLinux`).

vMohammad24:
> Internally, Home Manager has its own assertion to make sure the host isDarwin when you specify launchd options.

I did not know this was the case, LGTM.

Issue #1497, opened 2026-06-08
Localization


alba4k:
Vicinae has a lot of text elements that could be translated into other languages.

clankus-aurelius:
@aurelleb this issue may duplicate:

- #1704 — Both reports request translating Vicinae’s interface, with #1704 specifically requesting Simplified Chinese localization.

_Automated duplicate detection; maintainer confirmation is required._

aurelleb:
The localization infrastructure was implemented. For now we have support for French and Chinese.

alba4k:
Aight, omw to add italian to that list :)

alba4k:
@aurelleb 

<img width="1301" height="876" alt="Image" src="https://github.com/user-attachments/assets/a9ea242b-dfcc-4af6-9d12-067ce86322d4" />

are the numbers (e.g. `<location line="+10"/>`) safe to ignore? I'm guessing they just indicate where the translation is used

I tried using a few LLMs but they always failed in the worst ways (except maybe claude, which ran out of tokens so wasn't of much help either), so guess I'm doing it by hand 😭

aurelleb:
@alba4k sorry I missed this, the location lines are automatically generated when we run `make update-translations` so you don't need to maintain them or anything. They are used to show where the translation is used in the sources when using e.g QT Linguist.

FYI I used Fable 5 and gpt 5.5 pro to generate Chinese and French for the most part and it was okay

alba4k:
@aurelleb all good haha, figured it out

I tried gemini pro with max effort and it still crashed. but oh well, did it by hand and it is what it is. the pr is ready to imo. Been using it and have fixed the little "wrong" or strange translations I noticed

PR #1715, opened 2026-07-24
feat: localization support

aurelleb:
Adds the required infrastructure in order to support localization. We use the standard QT approach, where strings are marked for translation and translation catalogs get auto generated. Strings can then be edited using QT Linguist.

For now, localization only targets builtin Vicinae strings, and we give no way for extensions to localize themselves. Maybe it's something we will add in the future.

For now the intended way to contribute localization changes is to edit the catalogs using QT Linguist, but we may end up using something like [crowdin](https://crowdin.com/) eventually.

This PR also adds French language support as the first supported language.

(Most the changes in this PR are mechanical: most of them are about marking the strings for translations)


aurelleb:
fixes #1497

PR #1718, opened 2026-07-25
Add per-view window titles for KWin rules

zxsleebu:
### Motivation

Vicinae currently uses the same window identity for all launcher views. This means KWin window rules can only target the entire application, not individual views.

A practical example is Clipboard History: users may want to exclude it from screen capture without hiding every Vicinae window. To make that possible, KWin needs a stable property that uniquely identifies the active view.

### Changes

Views can now provide a stable window title suffix:

- Clipboard History: `Vicinae Launcher [clipboard-history]`
- Extensions: `Vicinae Launcher [extension:@author/extension:command]`

Other views keep the default title.

When switching views, the previous title is retained for two rendered frames to prevent sensitive content from briefly appearing
before KWin applies the new rule.


nolight132:
Works for me, waiting for it to get merged.

aurelleb:
I simplified the implementation and generalized it to every command instead of only handling it for a select few. The only difference is that the title format is now `Vicinae Launcher [<provider_id>:<command_id>]` . 
For the clipboard history, that's `Vicinae Launcher [clipboard:history]`

@nolight132 @zxsleebu please make sure everything still works for you and then we can merge this

zxsleebu:
Perfect! Works great!

PR #1721, opened 2026-07-25
feat(search): transliterate Cyrillic and Greek queries

nolight132:
Cyrillic and Greek queries now match Latin-titled items: `телеграм` finds
Telegram, `дискорд` finds Discord.

The table lives in `normalize.hpp` next to the diacritic folding it extends,
since one letter can need several ASCII chars (ж - zh). `queryVariants()`
returns the query plus its transliterations - at most 3, one per scheme.
`RootItemManager::search()` builds them once per search and keeps the best
score per item. The query stays first so same-script titles keep matching,
and ASCII queries get a single variant with identical scores.

Tested locally on Linux/Wayland.

image


aurelleb:
I modified the implementation a little so that it's not exclusive to root search and also doesn't over allocate memory.

Pretty cool improvement!

PR #1730, opened 2026-07-28
Feature: Add favorite toggle to commands in vicinae settings

seckinu:
Streamlines favoriting the commands you want to use through the settings menu rather than going one by one in the root search view

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -5,6 +5,6 @@
 # -DVICINAE_GIT_COMMIT_HASH=XXX (short rev)
 
 release:
-  tag: "v0.26.2"
-  rev: "1cf063c0b48a584ab6459394c51c9cfddd6acd26"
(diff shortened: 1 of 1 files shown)


src/server/src/config/config.hpp
```diff
@@ -25,6 +25,7 @@ struct ProviderItemData {
 
 struct ProviderData {
   std::optional<bool> enabled;
+  std::optional<bool> favorite;
```
aurelleb: Why do we need this? Setting a provider as favorite doesn't really mean anything at the moment

seckinu: @aurelleb I wasting testing stuff out to understand the codebase, must've forgot it, removed.

seckinu:
messed up the pr's, opened a new one at #1811 apologies for the confusion

PR #1734, opened 2026-07-28
MRU browser tabs

refractalize:
- sorts all browser tabs by MRU using their last access time
- puts tabs that are playing media into a top section, also MRU

image


aurelleb:
Really like the "playing media" tab
Thx

refractalize:
Thanks @aurelleb! Would you mind releasing the browser extensions so we can expose the tabs' lastAccessed times?

aurelleb:
oh yes I forgot, I will look into that

PR #1747, opened 2026-08-02
feat: allow excluding apps from the global toggle hotkey

voshond:
Some applications need to receive the global toggle hotkey themselves, but Vicinae currently always opens. This adds a per-application exclusion so the hotkey is ignored while an excluded application is focused.

### DIFF
--- a/.github/workflows/build-macos.yaml
+++ b/.github/workflows/build-macos.yaml
@@ -72,5 +72,6 @@ jobs:
         run: |
           ./build/bin/vicinae-glyph-tests
           ./build/bin/vicinae-fuzzy-tests
+          ./build/bin/vicinae-server-config-tests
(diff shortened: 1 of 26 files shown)


voshond:
As a honest note: i can somewhat follow the code, but i'm just a product manager, not an actual software engineer. This is 100% AI written & I have 0 C++ knowledge and i wouldn't be mad if you close this outright. Looking at the changes they seem reasonable to me, but i lack the experience to judge this PR with the scrutiny of someone who knows wtf they are doing, so i'm happy to keep changing this until someone is happy or have this as an inspiration for someone to implement this functionality in a proper way.

I did test this on archlinux with gnome 50 on wayland and macOS 15.7.7.

Either way, thank you for your time in advance.

aurelleb:
superseded by #1765

PR #1760, opened 2026-08-06
localization: Italian translations

alba4k:
As promised in https://github.com/vicinaehq/vicinae/issues/1497 haha

Mainly copy-pasted the `fr` file and edited from there, let me know if there's something that this approach won't cover

## Doubts regarding translations
* Is it fine to have `’%1’` instead of `&quot;%1&quot;`? was it left there as a copy paste mistake or does it have to be that way? Everything seems to work fine
* Should I translate "Are you sure?" as "Sei sicuro?" (masculine is sometimes used as neutral form, although it's uncommon/incorrect and can sound weird) or "Sei sicuro/a?" (masc/fem form)

## Strings that are currently not localized (or, at least, that weren't in `vicinae_fr.ts`)
* In the main launcher, pressing the vicinae logo (bottom left) shows "Extension Store"
* In the "Vicinae" settings screen, the list shows  shows "Extension Store" (probably just the above)
* In the Settings "About" Screen, "A focused launcher for your desktop -- native, fast, extensible"

## Weird thing that happen when not using english:
* In the "Applications" settings screen and others, the "Add Alias" button label is cut. I guess renaming it to just "Alias" or smth like that would work
* Checkmark out of bounds, see image below. The strings are being truncated too late

<details>

immagine

</details>

### DIFF
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -1219,6 +1219,7 @@ qt_add_translations(${TARGET}
 	TS_FILE_BASE vicinae
 	TS_FILES
 		${CMAKE_CURRENT_SOURCE_DIR}/translations/vicinae_fr.ts
+		${CMAKE_CURRENT_SOURCE_DIR}/translations/vicinae_it.ts
 		${CMAKE_CURRENT_SOURCE_DIR}/translations/vicinae_zh_CN.ts
 	PLURALS_TS_FILE ${CMAKE_CURRENT_SOURCE_DIR}/translations/vicinae_en.ts
 	SOURCES ${VICINAE_I18N_SOURCES}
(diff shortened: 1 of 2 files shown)


aurelleb:
Thank you for your work! 

> Is it fine to have ’%1’ instead of &quot;%1&quot;? was it left there as a copy paste mistake or does it have to be that way? Everything seems to work fine

Yeah I think so, if it shows correctly in QT Linguist then that's fine

> Should I translate "Are you sure?" as "Sei sicuro?" (masculine is sometimes used as neutral form, although it's uncommon/incorrect and can sound weird) or "Sei sicuro/a?" (masc/fem form)

I guess the latter then, but I don't speak Italian :p

About the UX issues, I will look into it thx

alba4k:
Would be the same as saying `Tu es très bel(le)` if I still remember some French. `sicuro/a` is what is already in the PR, let me know if anything else is needed on my end

aurelleb:
In French it's either `beau`, `bel`, or `belle` lol. Yes it's an annoying language.

alba4k:
I was decent in it in school. not fluent but could easily carry a conversation

but now I haven't needed it in a while so yeah..

PR #1765, opened 2026-08-08
feat: inhibit global shortcuts for specific frontmost apps

aurelleb:
Add a config file only setting to allow inhibition of global shortcuts when a specific app is in front.
Useful if dealing with a virtual machine or some sort of remote desktop software and you don't want vicinae to swallow keys.


voshond:
Tested, works for me :+1:

PR #1769, opened 2026-08-09
fix: execute action before closing window for Wayland XDG activation

### LINKED ISSUE #1697: [Hyprland] Sometimes a window will open on workspace 1 rather than the workspace I am on
**System information**

- Version: v0.23.2 (c55e85716)
- Build info: GCC 15.2.0 - Release - LTO
- Provenance: appimage
- OS: Debian GNU/Linux forky/sid - 
- QT Platform: wayland
- DE: Hyprland/Wayland

**Describe the bug**

Occasionally Vicinae will open new windows on workspace 1 rather than the workspace I am currently looking at.

**To Reproduce**

This is a hard one to reproduce. I tend to see it most with IntelliJ - but it doesn't always happen.

**Expected behavior**

To open on the workspace I am in.

**Screenshots**

N/A

**Additional context**

If useful, my settings file can be found here: https://github.com/markmandel/dotfiles/blob/main/provision/ui/files/config/vicinae/settings.json

markmandel:
Closing the launcher window before executing an action hides the window surface and resets QGuiApplication::focusWindow() to nullptr. As a result, when AppService attempts to request an XDG activation token, requestLaunchToken() lacks a focused window surface to bind to the activation token.

This seems to happen as a race condition that is particularly prevelent with applications that are slow to start (IntelliJ is a prime example).

Under compositors like Hyprland, an unauthenticated/unbound launch token fails to grant workspace focus to the newly spawned window, causing it to fall back to opening on workspace 1.

Executing the action prior to calling closeWindow() ensures the window surface and focus state remain valid when requesting the XDG activation token.

Closes #1697

### DIFF
--- a/src/server/src/navigation-controller.cpp
+++ b/src/server/src/navigation-controller.cpp
@@ -446,9 +446,8 @@ void NavigationController::executeActionNow(AbstractAction *action) {
     }
   }
 
-  if (action->autoClose()) { closeWindow({.clearRootSearch = true}); }
-
   action->execute(&m_ctx);
+  if (action->autoClose()) { closeWindow({.clearRootSearch = true}); }
   closeActionPanel();
 }
 


markmandel:
Worth noting - testing this out on my local machine, and so far, have yet to reproduce the issue again.

aurelleb:
I added xdg activation support very recently, in fact it is not in a release yet, so I doubt this is actually your issue 
Also, Hyprland doesn't care about xdg activation at all sadly (you need `focus_on_activate = true`)

That's a good fix nonetheless.

markmandel:
> I added xdg activation support very recently, in fact it is not in a release yet, so I doubt this is actually your issue

Ah, that's fair - I assumed it was the passing of the activation token, but that was my misunderstanding of how it works.

aurelleb:
It does just that but since this is very new it is unlikely to fix your issue (and as I said, Hyprland doesn't care about the token)

PR #1771, opened 2026-08-09
feat: add and migrate to figura beve generator

aurelleb:
Add a new code generator flag to figura to use the `beve` binary format instead of `json`.
This also adds a raw data type that can be used to transport data that should be parsed at a later point (e.g the manager service forwarding an extension message).

This PR also migrates the extension runtime and snippet IPC (on linux) to use that new `beve` generator, resulting in significant speedups.

This was made possible by the improved `beve` support that shipped with glaze `8.0.0`.


aurelleb:
actually this broke some stuff :(

PR #1792, opened 2026-08-13
fix(nix): copy Raycast extension output from $HOME, not /build

### LINKED ISSUE #1782: nix: mkRayCastExtension hardcodes /build in installPhase; fails on darwin with "cp: missing destination file operand"
## **System information**

Not a runtime bug: this is a packaging issue in the flake's Nix helper, so the in-app report-bug command doesn't apply. Observed on macOS (`aarch64-darwin`, Determinate Nix 2.34.8) with the vicinae flake at current `main`.

## **Describe the bug**

The default `installPhase` in [`nix/mkRayCastExtension.nix`](https://github.com/vicinaehq/vicinae/blob/main/nix/mkRayCastExtension.nix) copies the built extension out of a hardcoded `/build` path:

```nix
installPhase =
  attrs.installPhase or ''
    runHook preInstall

    mkdir -p $out
    cp -r /build/.config/raycast/extensions/${name}/* $out/

    runHook postInstall
  '';
```

`/build` is the Linux sandbox's build-directory convention, and it only works there indirectly: the nixpkgs npm hooks set `HOME` to the build directory (`export HOME="$TMPDIR"` in [`import-npm-lock`'s `npm-config-hook.sh`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/node/import-npm-lock/hooks/npm-config-hook.sh#L17)), `ray build` writes its output to `$HOME/.config/raycast/extensions/`, and on Linux `TMPDIR` happens to be `/build`.

On macOS there is no `/build`; with Determinate Nix, for example, build dirs live under `/nix/var/nix/builds/`. The hardcoded glob matches nothing, and since stdenv runs with `nullglob` the unmatched glob is dropped entirely, so the command degenerates to `cp -r "$out"/` and every Raycast-extension build fails with:

```
cp: missing destination file operand
```

## **To Reproduce**

On a darwin host, build any Raycast extension via the helper:

```nix
vicinae.lib.aarch64-darwin.mkRayCastExtension {
  name = "uuid-generator";
  rev = "<any raycast/extensions rev>";
  sha256 = "...";
}
```

`nix build` fails in `installPhase` as above. The same expression builds fine on Linux.

## **Expected behavior**

`mkRayCastExtension` builds on darwin as well as Linux, now that Vicinae itself supports macOS.

## **Additional context**

The portable fix is to copy from `$HOME` instead of hardcoding `/build`, since that's where `ray build` writes on every platform:

```nix
installPhase =
  attrs.installPhase or ''
    runHook preInstall

    mkdir -p "$out"
    cp -r "$HOME/.config/raycast/extensions/"*/. "$out"/

    runHook postInstall
  '';
```

I'm running exactly this today as an `overrideAttrs` wrapper around `mkRayCastExtension` in my home-manager config; with it the extensions (linear, spotify-player, tailscale, uuid-generator) build and run fine on `aarch64-darwin`, and behavior on Linux is unchanged (`$HOME` *is* `/build` there):

```nix
mkRaycastExt =
  args:
  (inputs.vicinae.lib.${system}.mkRayCastExtension args).overrideAttrs (_: {
    installPhase = ''
      runHook preInstall
      mkdir -p "$out"
      cp -r "$HOME/.config/raycast/extensions/"*/. "$out"/
      runHook postInstall
    '';
  });
```

(The `*/.` form also copies dotfiles and doesn't assume the directory `ray` creates matches `name` exactly; the minimal `s|/build|$HOME|` change works too.)

Happy to open a PR with this change if you agree with the approach.

mrbandler:
Fixes #1782.

`/build` only exists in the Linux sandbox. The npm hooks set `HOME` to the
build directory on every platform and `ray build` writes its output under
`$HOME/.config/raycast/extensions/`, so copy from there instead. On Linux
nothing changes (`$HOME` is `/build`); on darwin this fixes the
`cp: missing destination file operand` failure.

Tested on `aarch64-darwin` (Determinate Nix): building the `uuid-generator`
Raycast extension fails with the stock helper and succeeds with this change.

### DIFF
--- a/nix/mkRayCastExtension.nix
+++ b/nix/mkRayCastExtension.nix
@@ -41,7 +41,7 @@ lib.extendMkDerivation {
         runHook preInstall
 
         mkdir -p $out
-        cp -r /build/.config/raycast/extensions/${name}/* $out/
+        cp -r "$HOME/.config/raycast/extensions/"*/. $out/
 
         runHook postInstall
       '';


aurelleb:
thanks!

Issue #1790, opened 2026-08-13
Slow after macos sleep mode


ramlev:
**System information**

- Version: v0.25.0 (7e13b3f54)
- Build info: AppleClang 21.0.0.21000099 - Release - LTO
- Provenance: dmg
- OS: macOS Tahoe (26.6.1) (arm64)
- QT Platform: cocoa
- DE: Aqua

**Describe the bug**

When i hav used Vicinae for a day, and my MacMini M1 goes in lock mode for the night. When i unlock the computer the app is very slow at all tasks, opening the launcher takes 1-2 secs, slow when entering text in the launcher. Slow when i press the menu icon.

Only way to make it snappy again, is to quit and re-open

aurelleb:
@clankus-aurelius triage

aurelleb:
Should be fixed in v0.26.0, please let me know

ramlev:
after update launcher stopped  working and it interfer with other shortcuts (cmd+w) etc

aurelleb:
what do you mean stopped working? can you try to rebind the key to open it?

ramlev:
i have tried that, quit and repoen and re-bind the key to cmd+space, dont work, and neither does other cmd shorrrtcuts

aurelleb:
There was a problem indeed, please try again with release v0.26.2, I just released a fix

ramlev:
0.26.2 works :D Thank you for fixing fast, even without so much info from me :D

aurelleb:
You are welcome, let me know if your sleep issue is gone

ramlev:
It looks much better now, thank you

ramlev:
Actually after testing a few days with mac running 24/7 it's now slow again. Takes almost a second to open the launcher. And if i enter a string "php" it takes around 0.5-1 sec for every character to be deleted with backspace

Dissarion:
I have the same issue on release v0.26.3 (1f6de859a). It doesn't seem to be macOS itself, since I don't see any spikes in resource consumption, but the app becomes laggy. I can barely type anything in the search bar, or even open the search bar in the first place. It takes a few seconds for the app to load after I press the hotkey.

rscircus:
+1

rscircus:
> SIGABRT in malloc during QObject::~QObject → deleteChildren

[...]

PR #1796, opened 2026-08-14
refactor: remove tap-based macOS global shortcut backend

aurelleb:
Taps can be halted by macOS or slowed down by app nap, as such it is not robust enough to handle global shortcuts as a whole. We may reintroduce a tap if we want to implement something like an "hyper" key.


aurelleb:
fixes #1790

PR #1808, opened 2026-08-19
feat(glyph): add Emoji 17.0 emoji

itsameandrea:
Adds the eight new Emoji 17.0 (Unicode 17, September 2025) emoji to the picker database: 🫪 distorted face, 🫯 fight cloud, 🫈 hairy creature, 🧑‍🩰 ballet dancer (skinnable), 🫍 orca, 🛘 landslide, 🪊 trombone, 🪎 treasure chest.

`emoji-properties.cpp` already carries the Unicode 17 property ranges, so only `glyph.cpp` changes.

Since `gen.ts` isn't in the tree, the entries were produced with a throwaway script that follows the existing table conventions: ordering and grouping from UCD 17.0 `emoji-test.txt`, keywords from CLDR 48 `en` annotations plus name/subgroup words, new keywords appended at the tail of `kw_pool`, `g_index` re-sorted, section spans updated. As a safety check, the script's round-trip over the unmodified file is byte-identical after `clang-format`. Existing entries are untouched.

Tested locally: `make test` in `src/lib/glyph` passes (23 cases, 1988 assertions), plus a manual check that `glyph::lookup()` resolves all eight entries and `emoji::applySkinTone("🧑‍🩰", Medium)` yields the RGI sequence 🧑🏽‍🩰.


aurelleb:
superseded by #1812
The correct way was to regenerate using the fresh unicode data.

PR #1817, opened 2026-08-22
fix(macOS): index apps without bundle identifiers

gvanderclay:
## Summary

Some valid macOS application bundles do not have `CFBundleIdentifier`. Vicinae does not index these applications.

This change accepts a bundle only when macOS finds an executable file. Vicinae then uses the canonical bundle path as its internal identifier.

macOS opens the application from its bundle URL. Thus, a bundle identifier is not necessary for launch.

Vicinae uses the native bundle identifier only for running state, activation, window matching, and quit operations.

## Testing

- Ran `make check-format`.
- Built the macOS debug preset.
- Ran all macOS test programs.
- Started the patched build with an identifierless `Balatro.app`. Confirmed that the app appears in root search and launches.

AI assistance: OpenAI Codex was used to implement this change. I reviewed and tested the resulting code locally.


src/server/tests/app-service/macos/mac-app.cpp
```diff
@@ -0,0 +1,107 @@
+#include "services/app-service/macos/mac-app.hpp"
```
aurelleb: you can drop the tests for now, eventually we will do something more robust for testing these cases

aurelleb (changes requested):

src/server/tests/app-service/macos/mac-app.cpp
gvanderclay: Thanks for the quick response! Done! I dropped the tests and their CMake/CI setup.

aurelleb:
thanks!

PR #1824, opened 2026-08-23
fix(wayland): prevent settings window auto-close on focus handoff

### LINKED ISSUE #1822: Settings window closes immediately and crashes app (Wayland)
**System information**

- Version: v0.26.3 (1f6de859a)
- Build info: GCC 16.2.1 - Release - LTO
- Provenance: binary_release
- OS: EndeavourOS - 
- QT Platform: wayland
- DE: GNOME/Wayland

**Describe the bug**

On GNOME Wayland, the Settings window closes instantly after opening. Often, the entire app crashes.

**To Reproduce**

Steps to reproduce the behavior.

1. Use GNOME on Wayland with the Vicinae companion extension enabled.
2. Open the Vicinae launcher.
3. Press Ctrl + , or click the Settings icon.
4. The Settings window appears for a split second and immediately closes.

**Expected behavior**

The Settings window opens and stays open.

**Additional context**

The issue is caused by how the client and the GNOME extension handle Wayland surfaces and focus.

maxxxbond:
Fixes vicinaehq/vicinae#1822

This PR fixes the C++ side of the Wayland Settings crash.

Changes:

Uses QQuickWindow::closing instead of visibleChanged in settings-window.cpp. On Wayland, the visibility state toggles when a window maps. The old signal caused the window to delete itself instantly.
Waits for afterRendering on the Settings window before hiding the Launcher. The old code hid the Launcher too early and dropped focus to the desktop, which triggered the GNOME extension bug.
I will open a separate PR in the extension repository to fix the GNOME extension. Both PRs are needed to fully resolve the crash.

_Note: Code fixes were generated with the assistance of Gemini 3.7 Flash and Gemini 3.1 Pro._

### DIFF
--- a/src/server/src/actions/root-search/root-search-actions.cpp
+++ b/src/server/src/actions/root-search/root-search-actions.cpp
@@ -111,8 +111,8 @@ DisableItemAction::DisableItemAction(const EntrypointId &id)
 }
 
 void OpenItemPreferencesAction::execute(ApplicationContext *ctx) {
-  ctx->navigation->closeWindow();
   ctx->settings->openExtensionPreferences(m_id);
+  ctx->navigation->closeWindow();
 }
 
 void SetRootItemAliasAction::execute(ApplicationContext *ctx) {
--- a/src/server/src/actions/root-search/root-search-actions.cpp
+++ b/src/server/src/actions/root-search/root-search-actions.cpp
@@ -111,8 +111,8 @@ DisableItemAction::DisableItemAction(const EntrypointId &id)
 }
 
 void OpenItemPreferencesAction::execute(ApplicationContext *ctx) {
-  ctx->settings->openExtensionPreferences(m_id);
   ctx->navigation->closeWindow();
+  ctx->settings->openExtensionPreferences(m_id);
 }
 
 void SetRootItemAliasAction::execute(ApplicationContext *ctx) {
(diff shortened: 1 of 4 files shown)


aurelleb:
There is one problem with this, which is that since you no longer close the Vicinae window before showing the settings window, the settings window is not allowed to steal focus as the launcher window has exclusive keyboard interactivity (on layer shell configurations, Gnome being the exception)

In practical terms, this results in the settings window not getting its input focused as you open it (it might sometimes get focus, but that's a compositor race condition)

maxxxbond:
> There is one problem with this, which is that since you no longer close the Vicinae window before showing the settings window, the settings window is not allowed to steal focus as the launcher window has exclusive keyboard interactivity (on layer shell configurations, Gnome being the exception)
> 
> In practical terms, this results in the settings window not getting its input focused as you open it (it might sometimes get focus, but that's a compositor race condition)

Oh, I missed that since I was only testing on GNOME.

My GNOME extension PR (vicinaehq/gnome-extension#30) replaces the raw win.delete() call with a D-Bus signal, which fixes the focus drop crash on that side. Because of that, we don't need to delay closing the Launcher anymore.

I updated this PR to drop the afterRendering delay and restore the normal closing order.

The remaining code is still needed to fix the C++ side of the Wayland bug. visibleChanged is unreliable during window creation because compositors often toggle surface visibility during the initial mapping. The old code caught this toggle and instantly destroyed the Settings window before it even rendered.

Switching to QQuickWindow::closing fixes this by tying destruction strictly to actual close events.

Let me know if this is good to go

PR #1826, opened 2026-08-24
nix: add enableNumen option

vMohammad24:


vMohammad24:
@cswimr review if you have the time :)

nix/home-manager-module.nix
```diff
@@ -79,6 +79,15 @@ in {
       '';
     };
 
+    enableNumen = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
```
aurelleb: why `false`? Should be the new default

cswimr (approved):
LGTM.

PR #1828, opened 2026-08-24
feat(macOS): draggable window and snap guides

aurelleb:
Allow the launcher window to be freely moved on macOS, with snap guides to help placement.

https://github.com/user-attachments/assets/52502248-357a-4213-8396-f28927131534

### DIFF
--- a/src/server/src/qml/launcher-window-platform.hpp
+++ b/src/server/src/qml/launcher-window-platform.hpp
@@ -2,9 +2,17 @@
 #include <QtGlobal>
 
 class QQuickWindow;
+class QWindow;
 
 // Native windowing quirks the launcher window needs handled per platform.
 namespace LauncherWindowPlatform {
+
+#ifdef Q_OS_MACOS
+// Raises the drag snap overlay above the menu bar and Dock.
+void prepareOverlayWindow(QWindow *window);
+#else
+inline void prepareOverlayWindow(QWindow *) {}
+#endif
 
 #ifdef Q_OS_WIN
 // The system only grants foreground to the process that last received input;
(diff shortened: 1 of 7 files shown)


gvanderclay:
@aurelleb Ha, funny timing. I was working on almost the same feature and was just about to open a PR 😅

I pulled this down to try it and ran into a couple of edge cases I'd also hit in my version. With `"close_on_focus_loss": true`:

- A fast ⌘-drag can close the launcher when the cursor gets outside the window.
- Selecting text past the edge of the window closes it before you release the mouse.

They both seem to come from the same place. The out-of-bounds `MouseMove` handler closes the window even while a mouse button is held. This fixed both for me:

```cpp
if (me->buttons() == Qt::NoButton &&
    !contentRect.contains(me->position().toPoint())) {
  m_ctx.navigation->closeWindow();
}
```

Normal pointer-exit closing still worked after this change. Happy to send a small follow-up PR.

One other thought: I'd love an option to drag without holding a modifier. That's how Spotlight on macOS works, and it feels more natural to me. In my version I limited plain dragging to unused parts of the search bar so it wouldn't interfere with text selection. Maybe that behavior could be configurable eventually.

https://github.com/user-attachments/assets/83ce618e-1f56-4041-b978-9868ee42ff07

aurelleb:
ah, thanks! I forgot I implemented close on focus loss on macOS to be honest, as I don't use it. Indeed, that is not good. Gonna fix 

> I'd love an option to drag without holding a modifier. That's how Spotlight on macOS works, and it feels more natural to me. In my version I limited plain dragging to unused parts of the search bar so it wouldn't interfere with text selection. 

I tried it but found it too unreliable and unclear sometimes (+ adds additional maintenance to not conflict with future UI updates).

gvanderclay:
> I tried it but found it too unreliable and unclear sometimes (+ adds additional maintenance to not conflict with future UI updates).

That makes sense. I tried to brainstorm a way to allow for both, and the only idea I came up with was a drag handle. I'm not a huge fan of that since it feels like it could take away from the minimal aesthetic of the search bar. I can just retrain myself to drag with cmd click in this case!

PR #1835, opened 2026-08-24
feat(nav): ctrl/cmd + index shortcut for list views

### LINKED ISSUE #506: Alt+Number for quick entry selection
First of all, I really love this piece of software, by far my favorite Linux launcher.
I really like the possibility of moving through the different entries with CTLR+J/K but, coming from Flow Launcher, I still miss the possibility of quick selecting something with Alt+Number, more than a problem, it's just a personal commodity that I miss. 

A keybinding scheme or something similar to quick select entries with Alt+Number.

Moving with CTRL+J/K.

## **Additional context**
Flow Launcher example:

Image

### LINKED ISSUE #2000: Add shortcuts for launcher favorites and clipboard history
when I have pins / favorites elements, I can't press `cmd + 1`, `cmd + 2`, ... to select them, I have to click on the element to select it.

You should add something like I did on Tinycat in [PR 340](https://github.com/abue-ammar/tinycast/pull/340). It's written in swift but you get the idea : it should support multiple keyboard layouts and cross OS (`ctrl` and `cmd`).

The mapping uses hardware key codes rather than the character produced by the keyboard layout. This makes the shortcuts work consistently on layouts such as `QWERTY` and `AZERTY`.

You can also add a section to remap those shortcuts if you want to better support all keyboard layouts

## **Additional context**
This is what it looks like in Tinycast.

Image

PS: I mentioned the root palette and cipboard history but maybe it could be shared across all the launcher I don't know the other usages

tiagem:
### What changed
- Added a dedicated favorite action keybind (`Ctrl+Shift+F` by default).
- Wired `Ctrl+1`–`9` to launch the first nine favorites from root search.
- Show Command-held shortcut badges on favorite rows so the quick-open bindings are visible.

### Why
- This is a Raycast-style keyboard workflow that’s important for day-to-day launcher use.
- It was split out of #1806 so favorites reorder and shortcut handling can be reviewed separately.

Screenshot 2026-08-24 at 21 00 17

fixes #2000
fixes #506

### DIFF
--- a/src/server/src/extend/model-deser.cpp
+++ b/src/server/src/extend/model-deser.cpp
@@ -562,6 +562,7 @@ static const std::unordered_map<std::string, Keybind> NAMED_SHORTCUTS = {
     {"open", Keybind::OpenAction},
     {"open-with", Keybind::OpenAction},
     {"pin", Keybind::PinAction},
+    {"favorite", Keybind::FavoriteAction},
     {"refresh", Keybind::RefreshAction},
     {"remove", Keybind::RemoveAction},
     {"remove-all", Keybind::DangerousRemoveAction},
--- a/src/server/src/extend/model-deser.cpp
+++ b/src/server/src/extend/model-deser.cpp
@@ -562,7 +562,6 @@ static const std::unordered_map<std::string, Keybind> NAMED_SHORTCUTS = {
     {"open", Keybind::OpenAction},
     {"open-with", Keybind::OpenAction},
     {"pin", Keybind::PinAction},
-    {"favorite", Keybind::FavoriteAction},
     {"refresh", Keybind::RefreshAction},
     {"remove", Keybind::RemoveAction},
     {"remove-all", Keybind::DangerousRemoveAction},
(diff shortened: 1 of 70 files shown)


src/server/src/extend/model-deser.cpp
```diff
@@ -562,6 +562,7 @@ static const std::unordered_map<std::string, Keybind> NAMED_SHORTCUTS = {
     {"open", Keybind::OpenAction},
     {"open-with", Keybind::OpenAction},
     {"pin", Keybind::PinAction},
+    {"favorite", Keybind::FavoriteAction},
```
aurelleb: I don't think it makes sense to make this a named keybind, since it's only used in one place

aurelleb (changes requested):
I don't think it makes sense to scope that feature to favorites, this should work with every list item from index 0 to 9, in root search but also in any other vertical list view so that it works with extensions, clipboard history, etc...

src/server/src/extend/model-deser.cpp
tiagem: it makes sense what you saying, to be in all vertical lists. i'll update later

aurelleb:
fixes #2000
fixes #506

PR #1862, opened 2026-08-28
feat: add MPRIS now-playing media controls

mvanhorn:
## Walkthrough

HyperFrames walkthrough of this PR.

Walkthrough

[mp4](https://raw.githubusercontent.com/mvanhorn/vicinae/evidence/evidence/vicinae-1862-mpris.mp4)

Adds MPRIS media controls next to the existing volume commands: Play / Pause, Next Track, Previous Track (HUD like the volume ones) and a Now Playing list view.

`MediaControlService` mirrors `AudioControlService`: abstract backend + dummy + MPRIS backend. Linux only for now — macOS/Windows get the dummy and the extension is not registered there (macOS would need private MediaRemote APIs). No new QML, the view reuses `MonoListViewHost`.

Players are discovered on the session bus and the list is invalidated from `NameOwnerChanged` / `PropertiesChanged`, so nothing is polled while idle. Calls use a 1s timeout since players are third party programs, and a missing session bus simply yields no players.

An empty player argument acts on the last player controlled, falling back to the first playing one; otherwise it is fuzzy matched against the player name.

Tested with `-DBUILD_TESTS=ON` + `./build/bin/vicinae-server-tests`, and manually against a mock MPRIS player on a scratch session bus: discovery, `Identity`, the `xesam:artist` array, `PlaybackStatus`, transport calls, refresh on track change and on player exit, and no session bus at all.

Generated with Claude Opus 5.


mvanhorn:
HyperFrames walkthrough of this PR.

Walkthrough

[mp4](https://raw.githubusercontent.com/mvanhorn/vicinae/evidence/evidence/vicinae-1862-mpris.mp4)

aurelleb (changes requested):
Image

The icon to pause should be the pause one, not the play one, their usage is reversed here.

mvanhorn:
Good catch, fixed and pushed (also rebased onto current main).

`displayIcon` had the two states inverted: a playing track showed `PlayFilled` and a paused one showed `PauseFilled`. It now shows `PauseFilled` while playing and `PlayFilled` while paused, so the icon shows the action the row will perform.

Worth noting the action list right below it in the same file was already correct — `playing ? tr("Pause") : tr("Play")` with the matching icon — so the row and its own action were contradicting each other. They agree now.

I could not build locally to confirm visually (no Qt 6.9 on this machine), but the change is a two-value swap of existing `BuiltinIcon` members in an existing expression, so CI should cover it.

Issue #1628, opened 2026-07-08
Allow application search paths to be configured on macOS/Darwin


dnlzro:
## **Is your feature request related to a problem? Please describe.**

It's currently impossible to configure application search paths on macOS. They are hard-coded, and unlike on Linux, cannot be extended by appending paths to `XDG_DATA_DIRS`.

## **Describe the solution you'd like**

Either:

1. Add a new preference on macOS allowing users to specify which directories to search.
2. Read from `XDG_DATA_DIRS` on macOS as is already done on Linux.

## **Describe alternatives you've considered**

I tried to get around this by symlinking apps into one of the hardcoded paths (e.g., `$HOME/Applications`, but it appears that symlinks aren't followed.

## **Additional context**

This would make things easier for my nix-darwin setup, as I could add path entries from the Nix store (and thus avoid duplicating applications in one of the hard-coded paths).

hiSandog:
A native preference is less surprising than assigning Linux `XDG_DATA_DIRS` semantics on macOS, though the environment variable could still be an optional advanced input. Paths should be canonicalized and deduplicated, with a documented policy for following symlinked `.app` bundles from the Nix store and avoiding directory loops. An inspect/debug view listing each search root and discovered or rejected bundle would make custom-path problems much easier to diagnose.

PR #1868, opened 2026-08-30
fix(macOS): follow directory symlinks when scanning apps

dnlzro:
## Summary

A while ago I submitted #1628, mainly asking for configurable application search paths on macOS. But I also noted that symlinking apps into one of the default / hard-coded search paths (e.g. `~/Applications`) didn't work, since the app scanner doesn't follow symlinks. Later, #1777 added the configurable search-path preference (thank you!), but the symlink limitation still exists.

This is particularly relevant for nix-darwin setups, where applications live in the Nix store. By default, Home Manager *copies* the whole application bundle to `~/Applications/Home Manager`, but it's also possible to have them *symlinked* by specifying:

```nix
{
  targets.darwin = {
    copyApps.enable = false;
    linkApps.enable = true;
  };
}
```

My Macbook has a tiny SSD, so this makes a lot more sense than copying.

In short, this PR makes Vicinae follow the symlinks as expected.

- Apps inside symlinked subdirectories (of the configured search paths) now appear in results, where they were previously missed.
- Existing apps are unaffected.
- Deduplication of apps based on bundle IDs still applies.

## Main changes

1. `src/server/src/services/app-service/macos/mac-app-database.mm`

    `collectAppPaths()` now creates the `fs::recursive_directory_iterator` with the option  **`follow_directory_symlink`**, so symlinked directories under the search roots are descended into.

    This is safe with respect to symlink cycles because the existing traversal caps recursion at one level (`it.depth() >= 1 → it.disable_recursion_pending()`). Also, duplicated bundles discovered through overlapping symlinked paths are already deduplicated by bundle ID in `scan()`.

2. `src/server/src/services/app-service/macos/mac-app.mm`

    `MacApplication::iconUrl()` now resolves the bundle path with `std::filesystem::canonical` before handing it to `NSWorkspace iconForFile`. Without this, bundles reached through symlinks would render an icon with an ugly arrow in the corner (macOS uses this to denote symlinks in Finder, but once the application is launched, the normal icon appears in the Dock anyway). 

    Also, the resolved path is now cached on the instance (`m_iconPath`) so the syscall is made only once per app rather than on every lookup.

## Testing

I added a few tests in `src/server/tests/app-service/mac-app-database.cpp` and wired them into `make test` + macOS CI:

- Apps inside directories only reachable through a symlink are discovered by `scan()`.
- `scan()` terminates on symlink loops (self-referential symlink) and still finds apps.
- `iconUrl()` on a bundle reached through a symlink resolves to the canonical bundle path.

A minimal fake `.app` bundle (Info.plist + executable) is created in a temporary directory for each test.

Feel free to scrap these tests or refactor as you see fit! I wasn't sure exactly how to go about testing in this repo, this was just my best attempt.


aurelleb:
thanks!
Scrapped the tests for now, I need to come up with something a bit better to test this kind of stuff.

PR #1870, opened 2026-08-30
fix(calculator): map Qt locale names to valid C++ locales for Numen

### LINKED ISSUE #1869: Calculator: Numen backend (new default in v0.27.4) returns no result for any expression
**System information**

- Version: v0.27.4 (commit 47857b0b4)
- Build info: GCC 15.2.0 - Release - LTO
- Provenance: appimage
- OS: Omarchy (Arch Linux)
- QT Platform: wayland
- DE: Hyprland 0.56.2

**Describe the bug**

With the default `Numen` calculator backend, no calculator result appears in root search for any expression I tried. Typing `6+6` yields only the "Use "6+6" with…" fallback section and a Search Files row — there is no calculator row at all. `54 + 10%` behaves the same.

Selecting a different Calculator Backend (SoulverCore) makes results appear immediately, with no other change to the system. That isolates the failure to the Numen backend rather than to the calculator extension, the root search, or the install.

The failure is silent. The server logs `Started "Numen" calculator backend` at startup and never logs an error while queries are typed. The `calculator_history` table stays empty. Nothing indicates to the user that the backend is not producing answers.

Numen became the default in this release and only landed on 2026-08-22 (#1818), so this looks like a regression in a very new default rather than a long-standing issue.

**To Reproduce**

1. Install v0.27.4 (AppImage), leave Calculator Backend at its default (Numen)
2. Open the launcher and type `6+6`
3. No calculator row is shown

Then, for contrast:

4. Settings -> Extensions -> Calculator -> Calculator Backend -> SoulverCore
5. Type `6+6` again — the result now appears

**Expected behavior**

A calculator row showing `12`, as with the other backends.

**Additional context**

Because Numen is the default, a fresh install presents as "the calculator does not work" with no error anywhere to explain why. Even if the root cause is environment-specific, it would help a lot if a backend that returns no result for a well-formed expression logged something.

tiagem:
## Summary
- Numen was given `QLocale::name()` (`en_US`), which glibc rejects in favor of `en_US.UTF-8`, so every calculator query threw and was swallowed.
- Resolve a real `std::locale` name at backend start, keep Numen's error instead of a generic `"Error"`, and add a locale regression test.

Fixes https://github.com/vicinaehq/vicinae/issues/1869

### DIFF
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -362,6 +362,7 @@ set(SRCS
 	src/services/calculator-service/numen/numen-currency-provider.cpp
 	src/services/calculator-service/numen/numen-calculator-backend.hpp
 	src/services/calculator-service/numen/numen-calculator-backend.cpp
+	src/services/calculator-service/numen/numen-locale.hpp
 
 	src/services/files-service/abstract-file-indexer.hpp
 	src/services/files-service/dummy-file-indexer.hpp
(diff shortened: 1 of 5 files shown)


aurelleb:
fixed by #1871, the fix belongs in numen

PR #1872, opened 2026-08-30
feat(i18n): add German translations

ChrisLauinger77:
Introduces full German language support to the application.

- **Adds** the new `vicinae_de.ts` translation file, containing all necessary German strings for the user interface.
- **Updates** the CMake build configuration to include the newly added German translation, ensuring it is compiled and available at runtime.

This enhancement broadens the application's accessibility to German-speaking users.

**Breaking Changes:** None.
**Review Focus:**
- Verification of translation accuracy and contextual appropriateness within the application.


aurelleb:
thanks!

PR #1921, opened 2026-09-07
fix: look up unthemed app icons in XDG_DATA_HOME

shubham-cpp:
AppImage tools such as AppManager drop icons as loose files at `~/.local/share/icons/<name>.png`. Vicinae treats a desktop Icon= name as a themed icon via `QIcon::fromTheme`. The fallback search only walked **XDG_DATA_DIRS**, so those files were never found and search results showed the question-mark placeholder.

Use `commonDataDirs()` so fallback paths include **XDG_DATA_HOME**, and load an absolute Icon= value as a file, as the Desktop Entry spec requires.

| Before | After |
| --- | --- |
| PR for vicinae before fix | PR for vicinae after fix |

**Disclaimer**: Diagnosis and the patch were done with Grok 4.6(AI). I reviewed and tested the change.


aurelleb:
thanks!

PR #1933, opened 2026-09-10
fix(macOS): switch keyboard layout on capslock

### LINKED ISSUE #1694: Language switching via Caps Lock key does not work in search bar
## **System information**

- **OS:** macOS 26.5.1 (25F80)

## **Describe the bug**
When the macOS native keyboard setting "Use the Caps Lock key to switch to and from 'ABC'" is enabled, pressing the Caps Lock key inside the launcher's search bar does not switch the input language (e.g., between ABC and Russian). 

## **To Reproduce**
1. Open macOS **System Settings** > **Keyboard**.
2. Enable the option **"Use the Caps Lock key to switch to and from 'ABC'"**.
3. Open the Vicinae launcher and focus on the search bar.
4. Try to switch the input language by pressing the **Caps Lock** key.
5. Notice that the language does not change.

## **Expected behavior**
When this option is enabled in macOS settings, pressing the Caps Lock key while focusing on the search bar should successfully toggle the input language, matching native macOS behavior.

## **Screenshots**
–

## **Additional context**
If this option is disabled and the default **Fn** (Globe) key is used to change layouts, the language switches perfectly inside the search bar. The issue is specific to the Caps Lock switching mechanism.

DmitryTeplov182:
Fixes https://github.com/vicinaehq/vicinae/issues/1694

The search panel is intentionally a nonactivating `NSPanel` (`NSWindowStyleMaskNonactivatingPanel` + `_setPreventsActivation:`), so it never steals focus/activation from whatever app was in front — that's why Vicinae doesn't appear in Cmd+Tab or grab the menu bar when it opens. The previous discussion on this issue assumed Caps Lock language switching just isn't something Qt/the OS exposes to a non-native app, and that implementing it ourselves would be significant work.

That assumption turned out to be only half right. I verified on real hardware (macOS 26, Caps Lock configured to switch between ABC and a Cyrillic layout) with a live log capture:

- Physical Caps Lock presses *do* reach the app (`WindowServer: Caps Lock: false -> true`) regardless of activation state.
- `TextInputMenuAgent` (the system component that performs the actual switch for "Use the Caps Lock key to switch to and from ABC") only reacts when the *same* presses happen while a normal, activating window has focus — confirmed by testing in Vicinae's own Settings window (a regular window), where the shortcut works fine. It silently does nothing for the nonactivating search panel.
- Forcing the app to activate (`activateIgnoringOtherApps:`) while the panel was focused did **not** fix it either — so the gate is tied to the panel's nonactivating-panel bookkeeping specifically, not just app activation.

Since the OS won't perform the switch for us, the panel now does it itself: it watches for the Caps Lock `flagsChanged` event locally (this event *is* delivered to the panel, unlike the actual switch) and, when there's exactly one Latin and one non-Latin selectable keyboard layout enabled — the same condition macOS requires before it offers the preference in the first place — calls `TISSelectInputSource` to toggle between them, mirroring what the system would have done.

## Changes

- `macos-chrome-attached.{hpp,mm}`: the actual fix described above.
- `launcher-window.cpp`: unrelated crash fix hit while testing in dev mode — `unloadRoot()` nulls `m_window` before deleting the underlying `QQuickWindow`, whose teardown synchronously resigns key focus and re-enters the `activeChanged` lambda with the already-nulled pointer. One-line null check.
- `cli.cpp`: unrelated build-compat fix — `std::views::join_with` isn't implemented yet in the libc++ shipped with very recent Xcode, so `cli.cpp` failed to compile on that toolchain. Replaced with a plain loop.

## Testing

Built and ran locally on macOS 26.5.1 with Caps Lock configured for ABC/Russian switching. Verified: language switches correctly with Caps Lock while the search bar is focused, without affecting the panel's non-activating behavior (no focus stealing, no Cmd+Tab entry).

---
This fix was prepared with the help of Claude (Anthropic), who diagnosed the root cause via live log capture and a standalone repro on the reporter's machine before implementing the change.

### DIFF
--- a/src/cli/src/cli.cpp
+++ b/src/cli/src/cli.cpp
@@ -351,11 +351,8 @@ class DMenuCommand : public AbstractCommandLineCommand {
         {"index", ipc::DMenuOutputFormat::Index},
     };
 
-    std::string formatChoices;
-    for (const auto &pair : formatMap) {
-      if (!formatChoices.empty()) { formatChoices += ','; }
-      formatChoices += pair.first;
-    }
+    const auto formatChoices = formatMap | std::views::transform([](auto &&pair) { return pair.first; }) |
+                               std::views::join_with(',') | std::ranges::to<std::string>();
 
     app->add_option("-n,--navigation-title", m_req.navigationTitle, "Set the navigation title");
     app->add_option(
--- a/src/cli/src/cli.cpp
+++ b/src/cli/src/cli.cpp
@@ -351,8 +351,11 @@ class DMenuCommand : public AbstractCommandLineCommand {
         {"index", ipc::DMenuOutputFormat::Index},
     };
 
-    const auto formatChoices = formatMap | std::views::transform([](auto &&pair) { return pair.first; }) |
-                               std::views::join_with(',') | std::ranges::to<std::string>();
+    std::string formatChoices;
+    for (const auto &pair : formatMap) {
+      if (!formatChoices.empty()) { formatChoices += ','; }
+      formatChoices += pair.first;
+    }
 
     app->add_option("-n,--navigation-title", m_req.navigationTitle, "Set the navigation title");
     app->add_option(
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -732,6 +732,8 @@ if (APPLE)
 		src/services/file-chooser/macos/mac-file-chooser.mm
 		src/services/menu-bar/macos-menu-bar.hpp
 		src/services/menu-bar/macos-menu-bar.mm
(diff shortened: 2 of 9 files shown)


aurelleb:
Your fix wasn't quite it and it broke under several circumstances (switching spaces, repeted toggling, etc...). And in general, emulating this behavior like this is not a good solution.

I spent some time investigating and it turns out it's actually a QT bug that specifically affects non activating panels. QT already sets the activation policy and setting it again has no effect, so it needs to be set before the `QGuiApplication` is constructed.

For activating apps they still get the event because they are activated so it's doesn't seem to be an issue.

Might be worth a QT bug report

PR #1938, opened 2026-09-12
build(appimage): embed GitHub release update information

shubham-cpp:
## Intent

AppImage managers (AppManager, AppImageUpdate, Gear Lever) should pick up Vicinae's GitHub release URL automatically. Today they cannot, because the AppImage does not ship the metadata those tools look for.

## Problem

On install, AppManager fills **Update Link** / **Web Page** from:

1. the `.upd_info` ELF section (`gh-releases-zsync|owner|repo|latest|*.AppImage.zsync`)
2. `X-AppImage-Homepage` on the desktop file

We set neither. Tagged releases also do not publish a `.zsync` sidecar, so even a hand-pasted GitHub URL cannot do delta updates.

## Change

- `scripts/mkappimage.sh` tells linuxdeploy to embed `gh-releases-zsync|vicinaehq|vicinae|latest|*$ARCH.AppImage.zsync`. The repo is hardcoded so a rebuild from a fork still updates from official releases.
- `extra/vicinae.desktop` sets `X-AppImage-Homepage=https://github.com/vicinaehq/vicinae` (this is what AppManager shows as **Web Page**).
- Release and PR AppImage workflows install `zsync` (the pack image does not have `zsyncmake`) and upload `*.AppImage.zsync` next to the image.

After the next tagged release, installing the AppImage in AppManager should pre-fill Update Link and Web Page to `https://github.com/vicinaehq/vicinae`.

## Notes

I did not set `LINUXDEPLOY_OUTPUT_VERSION` / `LDAI_VERSION`. linuxdeploy would then rename the artifact from `Vicinae-x86_64.AppImage` to `Vicinae-<version>-x86_64.AppImage`, which would change the published asset names.

### DIFF
--- a/.github/workflows/build-appimage.yaml
+++ b/.github/workflows/build-appimage.yaml
@@ -59,6 +59,8 @@ jobs:
 
       - name: Make AppImage
         run: |
+          apt-get update
+          apt-get install -y zsync
           chmod +x ./scripts/mkappimage.sh
           ./scripts/mkappimage.sh "${INSTALL_DIR}" AppDir
 
@@ -66,6 +68,8 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: Vicinae-${{ matrix.target.deploy_arch }}-${{ github.sha }}.AppImage
-          path: "*.AppImage"
+          path: |
+            *.AppImage
+            *.AppImage.zsync
           if-no-files-found: error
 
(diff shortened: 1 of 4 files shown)


aurelleb:
As I mentionned in the issue I don't really want to encourage using the AppImage in a standalone way because some important things are set up by the script, and thus running it as an AppImage can lead to broken features, and more bug report noise.

shubham-cpp:
AppImage not being the recommended way to use vicinae is fair but the GitHub issue I've created is to add metadata information so that applications like [AppManager](https://github.com/kem-a/AppManager) can use that for auto-update functionality. 

So this is just about the metadata

PR #1952, opened 2026-09-14
fix(calculator): compute unpin position after erasing

BarganConstantin:
Unpinning a calculator history entry can write past the end of the record list.

`unpinRecord()` computes the position to re-insert the record, then erases the record, then inserts. The record being unpinned is always in front of that position, so the erase invalidates it. In the usual case (the entry is older than every unpinned one) the position is `end()`, and after the erase the insert writes one element past the end of the vector. Otherwise the record lands one slot too far down.

This erases first and then computes the position from the copied record.

Tested by running the old and new logic on the same list layout (pinned first, then unpinned by `created_at DESC`): the old one trips `_GLIBCXX_DEBUG` and AddressSanitizer reports a heap-buffer-overflow; the new one keeps the expected order for the only record, the oldest record, a record in the middle, and with another pin above it. The server builds with the change.


aurelleb:
thank you for the tiny, focused fixes. Much appreciated

BarganConstantin:
> thank you for the tiny, focused fixes. Much appreciated

Thanks! Glad it helped 🙂

PR #1975, opened 2026-09-16
fix(extensions): escape user text in generated package.json

BarganConstantin:
The "Create Extension" wizard writes the extension title/description/
author straight into package.json's JSON template with plain string
replace, and the form only checks minimum length. A description with
a quote or backslash (e.g. `A "simple" todo tracker`) produces a
package.json that isn't valid JSON, so the extension gets created but
then silently fails to load, with no error shown to the user.

Fixed by escaping backslash, double quotes, and newline/tab/carriage
return in each user-supplied field before substitution.

Tested by reproducing the exact template/substitution logic in a
standalone program linked against the real Qt6 JSON parser: the old
code produced "unterminated object", the new code parses cleanly.


aurelleb:
we probably need to replace this with glaze, I did that at the time because QJsonDocument doesn't have a way to preserve the order of the keys.

PR #1978, opened 2026-09-16
fix(numen): guard currency-fetch callback against dtor

BarganConstantin:
Symptom: quickly switching the Calculator Backend preference away from Numen and back while its currency-rate fetch is still in flight corrupts heap memory in vicinae-server.

Cause: NumenVicinaeCurrencyProvider isn't a QObject, so its QFutureWatcher::finished lambda captured raw `this` with no lifetime guard. Numen::setCurrencyProvider() destroys the old provider synchronously when the backend restarts, so a pending response later ran the lambda against freed memory.

Fix: add a shared_ptr<bool> alive flag, set false in the destructor, captured by the lambda; it bails out before touching `this` if the provider is already gone.

Tested with a standalone ASan repro of the exact capture pattern: the old shape reports heap-use-after-free, the new shape returns cleanly.


aurelleb:
it's probably cleaner to just make it a QObject imo

BarganConstantin:
Done — QObject now, so the connection dies with the provider and the flag is gone. Parented the watcher to it too, otherwise it leaks once the callback that deleted it stops running.

PR #1988, opened 2026-09-17
fix(macOS): derive node runtime and bundle slice from target arch

BarganConstantin:
Building the tree on an Intel Mac does not work: the node runtime is fetched for the wrong architecture, and `make mac-bundle` strips the binaries it just built.

Cause: the `APPLE` branch in `CMakeLists.txt` hardcodes `darwin-arm64`, and `scripts/macdeploy.sh` hardcodes `lipo -thin arm64`. The Windows and Linux branches directly below the first one already select by `CMAKE_SYSTEM_PROCESSOR`.

Fix: both now follow the target architecture — `CMAKE_OSX_ARCHITECTURES` when set, otherwise the host processor. On an Apple Silicon host the resolved values are identical to today's, so the arm64 build and the shipped dmg are unaffected. The `darwin-x64` checksum is the one published in nodejs.org's `SHASUMS256.txt` for the already-pinned `v22.23.1`.

This is not a request for Intel support or a CI runner. It only stops the tree from hardcoding an architecture where nothing else in it is arm64-specific: SoulverCore's xcframework already ships the `macos-arm64_x86_64` slice, and the official Qt builds are universal.

Tested on macOS 26.6.2, Intel Core i5-10600:

- against this branch: configure resolves the runtime to `node-v22.23.1-darwin-x64.tar.gz`, full build 428/428, `make mac-bundle` produces a bundle with no arm64 slices left, and the built binaries run.
- the same two changes applied to `v0.28.2` are installed and running as this machine's launcher.


aurelleb:
we don't support intel macs

BarganConstantin:
Understood, thanks for taking a look.

One clarification in case it was missed: this doesn't add Intel support and doesn't ask for a runner — on Apple Silicon it resolves to the same values, so the arm64 build and the published dmg are unaffected. If you'd still rather not carry the branching, that's fine by me.

aurelleb:
if we don't support it there is no need for the branching

PR #1992, opened 2026-09-18
fix: pass field value on form blur/focus for Raycast useForm

Meldiron:
## Summary
Raycast `@raycast/utils` `useForm` validates on blur via `event.target.value`. Vicinae notified blur/focus with empty args and `wrapFormItemProps` synthesized the value from React props only, so typed text was still treated as missing ("The item is required") and blocked submit.

- C++: `fieldBlurred` / `fieldFocused` now pass `item.effectiveValue()` (same pattern as `onChange`).
- TS: `wrapFormItemProps` prefers the native callback arg, falling back to `props.value` / `defaultValue`.
- Also route `PasswordField` through `wrapFormItemProps` like `TextField` / `TextArea`.

## Test plan
- [x] Google Meet extension → **Create Meet with Specified Profile** / **Add Profile**
- [x] Type into a required field, blur: error clears and typed value is kept
- [x] Submit succeeds when required fields are filled

## Manual QA

Before:

https://github.com/user-attachments/assets/d3eef7e3-9a9d-4441-b439-1fef3a981009

After:

https://github.com/user-attachments/assets/fd7de0dd-6538-47dd-8326-1bc67030112e


Meldiron:
Disclaimer: This is agentic contribution with Grok Bot. Problem found, and agent task written manually. Screen recordings of before and after were made manually too.

PR #2009, opened 2026-09-20
i18n: add Traditional Chinese (zh_TW) translation

ian902792:
Adds a zh_TW translation, wired into `qt_add_translations` the same way as the other locales.

I generated it by running the existing `vicinae_zh_CN.ts` through OpenCC (`s2twp` profile, which converts phrasing for Taiwan usage rather than just swapping characters), then went through the more common strings by hand to fix cases where simplified/traditional Chinese use different words for the same thing (e.g. file vs. document, "extension", settings-page wording).

I'm not a professional translator and this hasn't been proofread by another native speaker, so there are probably a few rough spots left, but it should already be solid enough to be useful and easy to iterate on from here. Happy to adjust anything maintainers want changed.

Built and ran it locally to make sure it actually loads and renders correctly before opening this.


aurelleb:
thanks!

Issue #91, opened 2025-08-20
feat: Add ignore app for clipboard history


cilginc:
I don’t want my passwords ending up in clipboard history, so I’d like to **ignore certain applications**, like my password manager.  

We could achieve this by checking `XDG_CONFIG_DIRS` and falling back to `~/.config/vicinae/clipboard`.  
Ignored applications could be listed in a YAML or JSON file, for example.

aurelleb:
One main issue I see is that we need to know which application triggered the selection. 

Currently, the `wlr-clipboard` implementation is unaware of this. As far as I know there is no proper way to get this info using the `wlr-data-control` protocol. The closest we could get would be to record the currently focused window when the selection occurs, but that's not very reliable.

Also the list of ignored apps would probably have to be managed using extension preferences instead of a config file, since this is the regular way we deal with application preferences. (this is also what Raycast does).

I don't know if this can help your use case, but we currently ignore selections that offer a `vicinae/concealed` mime type. This is mostly meant for internal usage (so that stuff copied from Vicinae is not stored in the clipboard history) but maybe it can be useful elsewhere.

Thinking of it, maybe your password manager offers a specific mime type that we could use to rule its selections out? I know some do.

ArjixWasTaken:
Password managers usually clear out the clipboard after a time interval, e.g. 5 seconds
Could that be detected?

Like, is the action of emptying the clipboard equivalent to writing empty data to it, or is it a separate action that can be detected?

aurelleb:
> Password managers usually clear out the clipboard after a time interval, e.g. 5 seconds Could that be detected?
> 
> Like, is the action of emptying the clipboard equivalent to writing empty data to it, or is it a separate action that can be detected?

a clipboard clear is a selection with no offers, I'm already ignoring these. But maybe there is something that can be done with that. Also that's not very reliable either.

cilginc:
> > Password managers usually clear out the clipboard after a time interval, e.g. 5 seconds Could that be detected?
> > Like, is the action of emptying the clipboard equivalent to writing empty data to it, or is it a separate action that can be detected?
> 
> a clipboard clear is a selection with no offers, I'm already ignoring these. But maybe there is something that can be done with that. Also that's not very reliable either.

Maybe path of the app can be usable. I know that you get the current window from desktop enviroment. Maybe that can be usable.

Hyprland have IPC support for example.

aurelleb:
[...]

PR #2013, opened 2026-09-21
feat(clipboard): ignore selections from specific apps

aurelleb:
Allow users to configure a list of applications whose selections will always be ignored.

When the clipboard server for the current platform is able to know what app the selection was obtained from, we use that. Otherwise, we use the current frontmost application, which is accurate most of the time (unless e.g the selection comes from some headless background process that doesn't own focus)

image


aurelleb:
fixes #91

